PR #2, opened 2021-04-24
Store protocols tokens value locked up in dynamodb

fabioDMFerreira:


src/storeTokensValueLockedInterval/computeTVL.ts
```diff
@@ -0,0 +1,270 @@
+import { StringNumber, Address } from "@defillama/sdk/build/types";
```
fabioDMFerreira, line 1: This function will be moved to https://github.com/DefiLlama/defillama-sdk.

src/storeTokensValueLockedInterval/getAndStoreTokensValueLocked.ts
```diff
@@ -0,0 +1,71 @@
+import { TokenPrices } from "../types";
+import { Protocol } from "../protocols/data";
+import storeNewTokensValueLocked from "./storeNewTokensValueLocked";
+import * as Sentry from "@sentry/serverless";
+import { TokensValueLocked } from "./tokensValueLocked";
+import computeTVL from "./computeTVL";
+
+
+export default async function(
```
fabioDMFerreira, line 9: This function will be moved to https://github.com/DefiLlama/defillama-sdk.

src/storeTokensValueLockedInterval/computeTVL.ts
```diff
@@ -0,0 +1,270 @@
...
+    }
+    Object.assign(tokenPrices, tempTokenPrices);
+    Object.entries(tempTokenPrices).forEach((tokenPrice) => {
+      knownTokenPrices[prefix + tokenPrice[0]] = tokenPrice[1] as any;
+    });
+  }
+  return tokenPrices;
+}
+
+export default async function (
```
fabioDMFerreira, line 81: This function will be moved to https://github.com/DefiLlama/defillama-sdk.

0xngmi:
Thanks for the PR! 

I changed computeTVLs.ts to provide the info you need on each token so I propose changing the PR to use that instead.

Also, we already compute all the TVL data in storeTvls calls so we could do everything on that lambda instead of adding another lambda that only gets the token data. That way we can save resources.

Apart from that everything looks good :llama:

fabioDMFerreira:
Using the same lambda and using the existing SDK compute TVL simplifies this PR code a lot. Thanks for the suggestions.

src/storeTvlInterval/storeNewTokensValueLocked.ts
```diff
@@ -0,0 +1,14 @@
+import dynamodb from "../utils/dynamodb";
+import { Protocol } from "../protocols/data";
+import { getTimestampAtStartOfDay } from "../utils/date";
+import { TokensValueLocked } from "../types";
+
+export default async (protocol: Protocol,
+  unixTimestamp: number,
+  tokensVL: TokensValueLocked) => {
+  await dynamodb.put({
+    PK: `dailyTokensVL#${protocol.id}`,
```
0xngmi, line 10: typo?

src/storeTvlInterval/getAndStoreTvl.ts
```diff
@@ -64,7 +63,12 @@ export async function storeTvl(
         continue;
       }
     }
-    await storeNewTvl(protocol, unixTimestamp, tvl);
+
+    const storeTokensAction = storeNewTokensValueLocked(protocol, unixTimestamp, tokensBalances);
```
0xngmi, line 67: This would be called every hour, but it seems that you'd like to store daily values using it, should it be changed? I think that storing both daily and hourly may be interesting since we may want to use that data in the future for charts with small timeframes

src/storeTvlInterval/storeNewTokensValueLocked.ts
fabioDMFerreira: what do you mean?

0xngmi: I was just wondering if `dailyTokensVL` should be `dailyTokensTVL`?

0xngmi:
Thanks!

PR #3, opened 2021-04-26
added slugs and removed some trailing spaces

returndiego:
The [`api.llama.fi/protocol/:name`](https://docs.llama.fi/api#get-protocol) endpoint of the defillama API is not ideal because it uses the display name (capitalized and with spaces) to query the database, for example, "Yearn Finance". 

The function behind this endpoint "slugifies" the query to find a match on the database.

https://github.com/DefiLlama/defillama-server/blob/662c27d8c395ed307278f240d4bc117e9c8db8f2/src/getProtocol.ts#L11

To avoid confusion and simplify the API usage, it'd be better to return a "slug" field, with the slugified version of the display name, for example: 

```json
{ 
...
"name": "Yearn Finance",
"slug": "yearn-finance"
}
```

on the [`api.llama.fi/protocols`](https://docs.llama.fi/api#get-all-protocols) endpoint, which can be then used to hit the Get Protocol endpoint: [`api.llama.fi/protocol/:slug`](https://docs.llama.fi/api#get-protocol).

If the change is made from `:name` to `:slug`, neither the server nor the API user have to manipulate the name field.

It took me a while to figure out how to use the Get Protocol endpoint, because the address bar automatically converts spaces into `%20` and I was getting no results. [See about spaces in urls here](https://stackoverflow.com/questions/5442658/spaces-in-urls).


returndiego:
The following protocols had trailing spaces on their names which I removed. 

https://api.llama.fi/protocol/synthetix
https://api.llama.fi/protocol/metronome
https://api.llama.fi/protocol/perlin
https://api.llama.fi/protocol/forcedao

These all return:

```
{"message":"Protocol is not on our database"}
```

Because of the issue that I raised in order to hit these endpoints you need to add a dash `-` at the end.

https://api.llama.fi/protocol/synthetix-
https://api.llama.fi/protocol/metronome-
https://api.llama.fi/protocol/perlin-
https://api.llama.fi/protocol/forcedao-

Please, change the API endpoint from `:name` to `:slug` in order to avoid these issues.

0xngmi:
Love this changes but I think it would be better to compute the slug for each protocol instead of hardcoding it

0xngmi:
I've re-implemented this PR by calculating the slugs instead. I believe this will make it easier to maintain.

Do the last changes solve your problem?

returndiego:
I think your solution is proper, thanks. 

I would just add a stronger slugify function, [like this one](https://gist.github.com/hagemann/382adfc57adbd5af078dc93feef01fe1). It should cover most possibilities. 

If you had use this one the trailing spaces wouldn't even matter. You won't have to worry about slugs again.

PR #6, opened 2021-05-07
Add script to get protocols old tvls

fabioDMFerreira:


0xngmi:
As discussed in another conversation, the problem with this PR is that it tried to pull old TVLs for all protocols (some of which don't support this) and that running this for all protocols would be too slow. In the meanwhile I've implemented this through other means so I'll just close this PR. Thanks for contributing :)

PR #17, opened 2021-06-25
Added missing types to the Protocol interface

diegoramosxyz:
I noticed protocol interface does not match what the API returns, so I added the missing types.

### DIFF
--- a/src/protocols/data.ts
+++ b/src/protocols/data.ts
@@ -1,4 +1,5 @@
 import { baseIconsUrl } from "../constants";
+import { chains } from '../types';
 
 export interface Protocol {
   id: string;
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -11,3 +11,20 @@ export type TokensValueLocked = {
 export type tvlsObject<T> = {
   [chain: string]: T;
 };
+
+const chains = [
```
0xngmi, line 15: There's more chains than the ones listed here and I expect this list to keep on growing so I think it may be best to forfeit some type safety in exchange for flexibility and not having to keep updating it all the time.

image

src/protocols/data.ts
```diff
@@ -14,9 +15,29 @@ export interface Protocol {
...
   cmcId: string;
   category: string;
-  chains: string[];
+  chains: chains[];
   module: string;
   twitter: string;
+  audit_links: string[];
+  slug: string;
+  tvl: number;
+  chainTvls: {
```
0xngmi, line 24: This type is not the one used to define the response returned from the API, but instead it's a type that defines the schema used internally to store the basic data. This data is then processed on https://github.com/DefiLlama/defillama-server/blob/master/src/getProtocols.ts, where we add the missing properties such as chainTvls, change_1h...

0xngmi (commented):
Thanks for the PR, left some comments

diegoramosxyz:
I'm closing this pr because I agree with your approach.

PR #19, opened 2021-07-15
fix /chain/Optimism

rajivpo:
This is an attempt to fix https://defillama.com/chain/Optimism (see image below for current state)

image

Didn't have env vars so was not able to validate locally. While it might also be a good idea to handle the case where no data is returned on the frontend, this would presumably fix the root cause.


0xngmi:
Thanks ser

PR #23, opened 2021-09-05
Adds Sunny Aggregator's CoinGecko market ID

SuryaKhosla:
See: https://www.coingecko.com/en/coins/sunny-aggregator


sunnymicrosystems:
👍

PR #24, opened 2021-09-08
Update kolibri data

Fitblip:
Right now you've got our gov token listed (kDAO), but kUSD is the actual stablecoin.

Let me know if you have any questions or want to discuss!


0xngmi:
yeah we list gov tokens instead of tokens used in the protocol because then we can do stuff like calculating mcap/tvl and it works better for the protocols that have multiple tokens

0xngmi:
Closing due to inactivity

PR #40, opened 2021-11-29
Order chains arr by tvl

shinitakunai:


src/getProtocols.ts
```diff
@@ -68,7 +68,7 @@ export async function craftProtocolsResponse(){
           slug: sluggify(protocol),
           tvl: lastHourlyRecord.tvl,
           chainTvls,
-          chains,
+          chains: chains.sort((b, a) => (chainTvls[b] - chainTvls[a])),
```
0xngmi, line 71: i believe some of these will undefined so sorting might not actually be correct

PR #67, opened 2021-12-05
update Platypus logo

shinji2:
update Platypus logo


0xngmi:
@realdealshaman could you update this on defillama app as well?

PR #233, opened 2022-03-14
add wormhole audits to data2.ts

justinschuldt:


justinschuldt:
Hey @0xngmi, here the audit doc https://github.com/certusone/wormhole/blob/dev.v2/audits/2021-01-10_neodyme.pdf

0xngmi:
applied this along with audit links on fa45938b

PR #322, opened 2022-04-28
change "Wormhole" to "Portal"

justinschuldt:
The TVL data supplied by this adapter is for the Portal bridge application specifically, which is built on the Wormhole protocol. The aim of this naming change is to be more clear about the source of the funds locked - these are funds locked in the Portal bridge contracts.

Want to be clear about the differentiation of TVL for the Portal bridge app, vs TVL of the protocol as a whole.

More info about Wormhole Protocol vs Portal Bridge:

https://wormholecrypto.medium.com/wormhole-protocol-vs-portal-bridge-whats-the-difference-a9ad98696e64 

https://twitter.com/wormholecrypto/status/1511822557854547969?s=20&t=6jT5LkUUAgarNIsnQoHALw


0xngmi:
ty, just wanted to make sure that it was correct since in the past we had scammers try to change pages, and for this case it wasn't fully clear

PR #482, opened 2022-06-26
Track FiatDAO audits

kargakis:


defi/src/protocols/data.ts
```diff
@@ -18568,7 +18568,12 @@ The eWIT token is a custodial, wrapped version of the Witnet coin managed by the
   description: "Fiat Protocol allows users to collateralize fixed income assets and mint FIAT with high loan-to-value ratios and low origination fees.",
   chain: "Ethereum",
   logo: `${baseIconsUrl}/fiat-dao.png`,
-  audits: "0",
+  audits: "3",
```
0xngmi, line 18571: should be 2 instead of 3

kargakis: updated

PR #501, opened 2022-07-02
update acryptos protocol adapters and chains

getunrekt:
+ update adapter for tvl data
+ update supported chain
+ update dapp URL


defi/src/protocols/data.ts
```diff
@@ -2016,18 +2016,18 @@ export default ([
...
     address: "bsc:0x4197c6ef3879a08cd51e5560da5064b773aa1d29",
     symbol: "ACS",
-    url: "https://app.acryptos.com/",
+    url: "https://app-v2.acryptos.com/",
     description:
       "Advanced Crypto Strategies, Yield Optimizer and StableSwap on BSC.",
     chain: "Binance",
     logo: `${baseIconsUrl}/acryptos.jpg`,
-    audits: "2",
+    audits: "6",
```
0xngmi, line 2024: this should be kept at 2, its not a number of audits

getunrekt: thx for highlighting.
i've just push a fix , setting it back to 2.

PR #544, opened 2022-07-13
Coins

waynebruce0x:


coins/serverless.yml
```diff
@@ -139,6 +141,8 @@ functions:
   fetchHourlyCoingeckoData:
     handler: src/fetchCoingeckoData.fetchHourlyCoingeckoData
     timeout: 900
+  storeDefiCoins:
+    handler: src/storeCoins.storeCoins
```
0xngmi, line 145: need cron trigger?

coins/src/storeToDB/adapters.ts
```diff
@@ -0,0 +1,3 @@
+export default {
+  aave: require("../adapters/moneyMarkets/aave/aave.ts")
```
0xngmi, line 2: why not use import and get free type checks?

coins/src/storeCoins.ts
```diff
@@ -1,31 +1,17 @@
...
-} from "../src/utils/shared/coingeckoLocks";
-import protocols from "./protocols/data";
-import { importAdapter } from "./protocols/importAdapter";
-
-const maxRetries = 4;
+import { importAdapter } from "./storeToDB/importAdapter";
+import protocols from "./storeToDB/data";
 
 async function iterateProtocols(protocolIndexes: number[]) {
   const actions = protocolIndexes
```
0xngmi, line 5: is the idea here that multiple protocols will use the same adapter? if that were to happen wouldn't it be easier to store those protocols inside the adapter?

coins/serverless.yml
```diff
@@ -66,6 +66,7 @@ provider:
     HARMONY_RPC: ${file(./env.js):HARMONY_RPC}
     DEFILLAMA_SDK_MUTED: true
     tableName: ${self:custom.tableName}
+    AWS_REGION: ${self:custom.region}
```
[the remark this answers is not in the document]
waynebruce0x: batchGet() doesnt seem to work without it

coins/src/adapters/lps/uniswap/uniswap.ts
```diff
@@ -1,83 +1,210 @@
...
+    );
+  }
+  return addresses;
+}
+async function fetchUniV2MarketData(chain: string, pairAddresses: string[]) {
+  let token0s: multiCallResults;
+  let token1s: multiCallResults;
+  let reserves: multiCallResults;
+  [token0s, token1s, reserves] = await Promise.all([
+    multiCall({
```
0xngmi, line 67: what about setting requery:true, instead of calling requery afterwards?

in this case we can do that since we know for sure that all calls must be successful

coins/src/adapters/moneyMarkets/aave/aave.ts
```diff
@@ -56,23 +66,24 @@ async function getTokenPrices(chain: string, registry: string) {
...
-      tokenInfo.decimals[i].output,
-      tokenInfo.symbols[i].output,
-      underlyingRedirects.filter((u) =>
-        u.PK.includes(r.input.params[0].toLowerCase())
-      )[0].redirect
-    )
-  );
+  reserveData.map((r, i) => {
+    try {
+      addToDBWritesList(
```
0xngmi, line 71: why would this fail?

coins/src/adapters/moneyMarkets/aave/index.ts
```diff
@@ -0,0 +1,34 @@
...
+      "v2"
+    ),
+    // AMM market has no registry
+    //getTokenPrices("ethereum", "0x7937d4799803fbbe595ed57278bc4ca21f3bffcb");
+    getTokenPrices(
+      "polygon",
+      "0x3ac4e9aa29940770aeC38fe853a4bbabb2dA9C19",
+      "v2"
+    ),
+    getTokenPrices("avax", "0x4235E22d9C3f28DCDA82b58276cb6370B01265C2", "v2")
```
0xngmi, line 27: what about adding the other v3 deployments?

coins/src/adapters/moneyMarkets/compound/index.ts
```diff
@@ -0,0 +1,7 @@
+import getTokenPrices from "./compound";
+
+export function compound() {
+  return Promise.all([
+    getTokenPrices("ethereum", "0x3d9819210a31b4961b30ef54be2aed79b9c9cd3b")
```
0xngmi, line 5: could add a bunch of compound forks as well? maybe under different adapters?

0xngmi (approved):
looks pretty good, i guess only thing missing is how we can get historical prices?

coins/src/adapters/moneyMarkets/aave/aave.ts
waynebruce0x: because all the atokens redirect to the underlying, if the underlying isnt in the DB we cant direct it anywhere, so it'll error on finding 'redirect' on undefined

waynebruce0x: its not an issue on aave but I guess it could be on forks

PR #554, opened 2022-07-15
change ALEX category

hyossid:
Hi, since our platform is not limited to Yield, we would like to change the category name as Defi.


0xngmi:
defi as a category doesnt exist, pick a category that already exists

hyossid:
> defi as a category doesnt exist, pick a category that already exists

Changed to Dexes

PR #600, opened 2022-07-25
Update Beefy

MirthFutures:
Updating Beefy's description, name, web link and chains.


0xngmi:
@realdealshaman can you verify

PR #670, opened 2022-08-04
Remove pool token from description

kargakis:
BCT is just one of the pool tokens that
we have enabled to be pooled but should
not be associated as our main protocol
token.


kargakis:
cc @Define101 @g1nt0ki

PR #814, opened 2022-08-29
fixing tifi url and desc

007vasy:
fix tifi desc/url error


007vasy:
@0xngmi , @Nemusonaneko, @realdealshaman

PR #889, opened 2022-09-12
Update data for Redacted Cartel

justingreenberg:


Define101:
Hello guys 
Could I suggest a different approach? 
I notice that redacted cartel have different products 
Could be better to do something as we have with AAVE protocol and other protocols, basically having Redact on the left side and all the projects from redact on the right side. 
To make this, we would need an adapter to count the TVL on Pirex Protocol and Hidden Hand
Since the current adapter is counting the TVL from redacted Protocol
image

image

Define101:
Hello Justin
I am confirming with the team on what should be the best approach in this case
But kinda agree with your second option

Define101:
Hello guys
I have made the changes 
So Redacted will be the parentProtocol and will have redacted protocol and Pirex convex under 
Once you guys added more things we can change it back to Pirex and add all the new things under the same adapter

Regarding the 2 Point. 
After speaking with the team seems that Hidden Hand would fit on the fees adapter (we just launched yesterday) https://defillama.com/fees
Just confirming the best approach to add a Fees Adapter without having a TVL adapter, since hands have no TVL

swheel33:
Congrats on the launch and thanks for making the updates! I agree with you I think Hidden Hand fits great with the fees adapter. I was looking on the docs and there didn't seem to be a page for setting that up, what's do you need from us in order to get listed?

Define101:
Yeah launch our fees adapter 1/2 days ago, so it's very early, there is still a bit of work to do updating the docs
Would recommend checking out our discord , someone will help you out

Define101:
Hi guys, we just updated our fees docs yesterday
Take a look at https://docs.llama.fi/list-your-project/how-to-add-a-fee-adapter
Can I close this PR?

swheel33:
Thanks! Yea you can close it I'll open another one for the Hidden Hand fees

PR #921, opened 2022-09-18
Update thorchain website

dtmkeng:
Update thorchain website


Define101:
Thanks, seems that the .org is taking too much to load the page
Since the .com is also an official link, we will change it to that

dtmkeng:
ty

PR #1101, opened 2022-10-13
track more info on coins

0xngmi:


0xngmi:
will merge tomorrow so i can monitor if anything breaks

0xngmi:
wtf this is so old, i dont remember why i didnt merge

PR #1497, opened 2022-11-28
feat: add values for across token

abg4:
Signed-off-by: Alex Gaines <againes@umaproject.org>

Added values for ACX token based on the launch earlier today.


Define101:
Hi guys , i am gonna make the changes
There is only one problem, the adapter currently is only counting the ETH network 
Would you guys be able to update the code to count the missing chains? "Polygon", "Arbitrum", "Optimism", "Boba"

Define101:
I have added all the new information (just not the chains). Also if you guys could provide us with some contract addresses for those chains would be great 
Should be live within 1 hour

PR #1498, opened 2022-11-28
Fix: update Outcome to remove UMA token

abg4:
Signed-off-by: Alex Gaines <againes@umaproject.org>

This PR removes UMA token values that should not be included with Outcome.


Define101:
Hi, just made the changes
The same problem as the other PR, our adapter for outcome currently only supports eth, polygon, and boba network
If you guys could updated the current adapter to support those new chains would be great

PR #1522, opened 2022-12-01
feat: add mean finance ethereum

0xsambugs:
We are now adding support for ethereum tokens, for mean finance


coins/src/adapters/yield/mean-finance/mean-finance.ts
```diff
@@ -5,7 +5,7 @@ import { fetch } from "../../utils";
 
 export default async function getTokenPrices(chain: string, timestamp: number) {
   const tokens: { type: string; address: string }[] = await fetch(
-    `https://api.mean.finance/v1/dca/networks/${chain}/tokens`
+    `https://api.mean.finance/v1/dca/networks/${chain}/tokens?includeNotAllowed`
```
0xsambugs, line 8: This is so that we can include tokens that were allowed at some point, but can't be deposited at the moment. Even if that is the case, it makes sense to price them and mark them as TVL

PR #1564, opened 2022-12-09
Update Cozy Finance for UMA as oracle

abg4:
Cozy uses UMA as an oracle. This PR adds Cozy to UMA's TVS.


Define101:
Hi abg4 , after reading their docs I can only find a reference to chainlink 
Do you have any proof they are using UMA oracle?
image
https://docs.cozy.finance/what-is-cozy/faq

abg4:
I should have given more details around the integration. Cozy uses UMA for their V2 which they just announced:
- https://twitter.com/cozyfinance/status/1600901786658693120?s=46&t=u_yAqfcUgFesJ767twfD3w
- https://twitter.com/hal2001/status/1602122062621208578

Here is an Optimism contract requesting a price from the UMA oracle with the description, `"Pays out if the UMA oracle determines there was a hack or exploit on dYdX resulting in loss of funds."`
- https://optimistic.etherscan.io/tx/0x790903cfad0ce93ce4ba7606d76149e2b10d9b8ea1b163ebad589cd9b8f99a9e

Let me know your thoughts/questions or if you think we should wait to add until there is more info on the integration.

PR #1716, opened 2022-12-24
update desc & add gecko_id for Hedgehog

ivanvolov:


Define101:
I have made the changes + the new logo
Should be live within 1 hour

PR #1847, opened 2023-01-06
fix: Metavault.Trade Protocol Info

0xrmvdao:
We've updated Metavault.Trade protocol info. Metavault.Trade made update on short and long leverage from 30x to 50x now.


Define101:
Thanks for the PR 
Have a great day

PR #1905, opened 2023-01-13
Add FiRM protocol by Inverse Finance

webmass:


Define101:
Thanks for your PR , we have listed Inverse Finance Firm protocol
However regarding changing Inverse Finance name, we would prefer to keep the same name and instead, add a hallmark inside the adapter (we will do it)

Define101:
Could you tell me when was the project deprecated?

webmass:
Hey,

Thanks for the review

Since june 16th 2021 Frontier can be considered deprecated as borrowing was disabled permanently.

I guess the best would be something like the Aave page https://defillama.com/protocol/aave where there is a tab system per protocol with each tab having its own data and infos, in that case FiRM should be the default tab as it's the only active protocol.

Define101:
Hello webmass, sorry for the late reply.
First of all , thanks for your fast reply
Regarding the parentProtocol would fit well here (what we have with aave).
So we will call the parentProtocol Inverse Finance 
The current Inverse FInance (deprecated) will be called Inverse Finance Frontier ( we will also add a hallmark informing about the deprecation)
Will let you know once is ready

Define101:
Should be live within 1 hour. 
Gonna keep the PR open for the next 24h, if there is anything else just let us know

Define101:
https://defillama.com/protocol/inverse-finance

webmass:
Hey,
Sorry I made a typo regarding the deprecated date, it's june 16th 2022 (not 2021).

The deprecated hallmark is nice, though I'm thinking it would be nice to have this small feature:
- `deprecated` as a protocol attribute
- if true its shows a "deprecated" badge top-right of the title in lists/tabs (so a user knows it's deprecated without having to click on it)

webmass:
Just noticed a copy-pasta issue:
FiRM protocol is not forked from anything.

Shouldn't FiRM protocol show "audited" yes with the audit note?

Define101:
image
Not sure if we can consider that an audit, we have removed the compound forked.
Regarding the Feature, we will discuss it internally, will be back to you once we have more informations

webmass:
Thanks, I see

Regarding the audit part, I agree that `code4rena bug bounty contests` are somewhat different than standard audits, but one can argue that it's actually  a better form of audit as there's much more reviewers involved, I'd say it at least needs its own "audit category", meaning showing "Code4rena bug bounty contest" or "bug bounty contest" for example ^^.

Define101:
Hello webmass, after internally discussing with the team, we cant add code4arena as an audit 
Hope you understand

webmass:
> Hello webmass, after internally discussing with the team, we cant add code4arena as an audit Hope you understand

Hey, 
it's fine to not see code4rena as a standard audit, but I also think in this case that it makes sense to add a different field for bug bounty contests and simply display "Bug Bounty Contest: yes+link / hidden field if none", more and more projects use code4rena or similar as this is a more decentralized approach, which is a better fit with the crypto-ethos than old-school "corporate auditors".

g1nt0ki:
> > Hello webmass, after internally discussing with the team, we cant add code4arena as an audit Hope you understand
> 
> Hey, it's fine to not see code4rena as a standard audit, but I also think in this case that it makes sense to add a different field for bug bounty contests and simply display "Bug Bounty Contest: yes+link / hidden field if none", more and more projects use code4rena or similar as this is a more decentralized approach, which is a better fit with the crypto-ethos than old-school "corporate auditors".

closing this PR for now.
Your suggestion is good, I would wait for more projects to request for the same before adding a new field

g1nt0ki:
FiRM was added in this commit: https://github.com/DefiLlama/defillama-server/commit/92335c9af4fef5ef928c023a5f81f2365d2d81cb

PR #2118, opened 2023-02-06
Rename "Osmosis" protocol to "Osmosis DEX"

sunnya97:
This is to differentiate the dex protocol from the Osmosis chain as a whole


Define101:
Thanks for this PR
We have made the changes and should reflect within 1 hour
Will close this PR

PR #2232, opened 2023-02-20
emissions chart endpoint WIP

waynebruce0x:


defi/serverless.yml
```diff
@@ -569,6 +569,13 @@ functions:
       - http:
           path: npm/{start}
           method: get
+  emissions:
+    handler: src/getEmissions.default
+    timeout: 900
```
0xngmi, line 574: max timeout for lambdas that get called from api gateway is 30s

PR #2249, opened 2023-02-21
Remove parentProtocol from Ondo V1

nickplace:


realdealshaman:
@nickplace we would prefer to keep it for historical tvl

nickplace:
Created a more thorough PR with better explination

PR #2298, opened 2023-02-24
chore: Add dexter as protocol

abhimanyusinghgaur:


realdealshaman:
Hi @abhimanyusinghgaur  , thank you for the PR, it's much appreciated. Having this info is definitely helpful. We will add it to the file once we have an adapter for Dexter, but it looks like it's currently only available on testnet. Could you please let us know when it goes live so that we can look into pulling its TVL? Thank you again

Define101:
https://github.com/DefiLlama/defillama-server/pull/2680

Define101:
Should be live within 1 hour.

Define101:
https://defillama.com/protocol/dexter

PR #2301, opened 2023-02-24
Update parcl protocol data

DMJ16:


0xngmi:
merged manually in 216a6fa6

PR #2522, opened 2023-03-14
Fix issue associating Ondo Finance v1 and Ondo Finance

nickplace:
Ondo v1 and Ondo Finance are completely separate endeavors from different organizations. Ondo v1 was a DeFi-native protocol supported by an offshore (Cayman) foundation. Ondo Finance, a US-based entity, has separately launched centralized regulated security tokens like OUSG. We are supportive of preserving the historical record related to Ondo v1 through its own page in defillama but need to avoid confusion with the regulated products issued out of Ondo Finance today.

### DIFF
--- a/defi/src/protocols/data.ts
+++ b/defi/src/protocols/data.ts
@@ -7313,7 +7313,7 @@ The eWIT token is a custodial, wrapped version of the Witnet coin managed by the
   },
   {
     id: "460",
-    name: "Ondo Finance V1",
+    name: "Ondo v1 (Legacy)",
     address: "-",
     symbol: "-",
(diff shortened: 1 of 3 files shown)


0xngmi:
isn't the ondo token related to both?

nathanallman:
The ONDO token is not directly related to either. It's actually the governance token for the Ondo DAO which is currently related only to Flux Finance, somewhat confusingly. We may make and support some naming adjustments in the future to clarify these distinctions but for now would really prefer to keep these distinctions clear.

nickplace:
@0xngmi thoughts?

PR #2533, opened 2023-03-15
Block sanity check

waynebruce0x:


coins/src/getBlock.ts
```diff
@@ -46,6 +46,20 @@ function getExtraProvider(chain: string | undefined) {
   return getProvider(chain as any);
 }
 
+async function isAValidBlockAtThisTimestamp(timestamp: number, provider: any) {
+  try {
+    const [genesisBlockTime, latestBlockTime] = await Promise.all([
+      provider.getBlock(1).then((b: any) => b.timestamp),
```
0xStrobe, line 52: smol thots -

1. should we cache the genesis block time so we reduce rpc calls? its static
2. should we use "now" instead of latestBlockTime? latest block time will always be in the past and if we want to check the closest block to "now", this function will say it's an invalid timestamp right?

PR #2535, opened 2023-03-15
add RealT tokens

mastilver:
They only have the latest price using this API, I don't know if it's fine.
I set the confidence to 1, the price from the API is the price they are selling it to you for and they will buy it back from you with a 3% fee.

### DIFF
--- a/coins/src/adapters/other/index.ts
+++ b/coins/src/adapters/other/index.ts
@@ -5,6 +5,7 @@ import unknownTokenAdapter from "./unknownToken";
 import podsAdapter from "./pods";
 import distressedAdapter from "./distressedAssets";
 import manualInputAdapter from "./manualInput";
+import realtAdapter from "./realt";
 import { wrappedGasTokens } from "../utils/gasTokens";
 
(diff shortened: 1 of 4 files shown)


waynebruce0x:
looks great. I'm tempted to set the price to the buy back price. My logic is that this is the price 'circulating' tokens are valued at. This is assuming that realT is the only available market. What do you think @mastilver ?

coins/src/adapters/other/index.ts
```diff
@@ -73,3 +74,10 @@ export function manualInput(timestamp: number = 0) {
     manualInputAdapter("polygon", timestamp),
   ]);
 }
+export function realt(timestamp: number = 0) {
+  console.log("starting realt");
+  return Promise.all([
+    realtAdapter("ethereum", timestamp),
+    realtAdapter("gnosis", timestamp),
```
0xngmi, line 81: chain name might need to be xdai instead

mastilver: fixed :)

Sigri44:
Look for our Community API : https://api.realt.community/

PR #2651, opened 2023-03-26
Update data2.ts

covofin:
Update Info for COVO for Protocol Upgrade

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -24678,7 +24678,7 @@ const data2: Protocol[] = [
     symbol: "COVO",
     url: "https://covo.finance",
     description:
-      "Covo Finance is a decentralized trading platform that offers spot and perpetual trading with low swap fees and zero market impact. The platform offers a maximum leverage of 50x on major cryptos and is backed by a singular multi-asset pool that generates income for liquidity providers through fees garnered from market making, swap transactions, leveraged trading (including spreads, funding fees, and liquidations), and rebalancing of assets",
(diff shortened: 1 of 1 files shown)


Define101:
Why remove Fork information?

covofin:
We have upgraded our protocol to version 2, which is no longer a GMX fork and allows trading of even Forex and Metals, and we are completely deprecating version 1 and focus only on V2. The V2 adapter has already been approved and can be found at https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/covo-v2/index.js."

covofin:
The V2 platform is totally different from GMX in terms of contracts and user interface. Unlike GMX, where all listed assets require a certain percentage of allocation in the pool like BTC, ETH, etc., On our V2 platform all trades are backed by the single USDC pool, which allows for leverage trading of any listed assets including Forex pairs like USD and EUR through Pyth and Chainlink oracles. This capability is not available on GMX, thus we are no longer a fork of GMX.

Define101:
We have created a parentProtocol
Now we track V1 and V2 in different listings
I will keep GMX fork on Covo V1, and remove the fork information from Covo V2
Is this ok ?

covofin:
Okay, this can work for V2.

covofin:
Also, for Fees and Revenue tracking, can we somehow mention that current stats are for V1 only, as we are currently working on adaptors to track V2 Fees and revenue.

Define101:
"can we somehow mention that current stats are for V1 only" We will point to Covo V1
Once there is an adapter for the v2 , there will be covo v1 and covo v2 on the fees dashboard

PR #2655, opened 2023-03-27
add Synthetix under Pyth TVS

mariobern:
gm llama sers

would like your help to review Pyth as another oracle for Synthetix

They have been a strong supporter since late last year and we want to add them under Pyth TVS

https://twitter.com/PythNetwork/status/1595099857231831041?s=20

https://twitter.com/synthetix_io/status/1601280873965371392?s=20

Thank you sers

### DIFF
--- a/defi/src/protocols/data.ts
+++ b/defi/src/protocols/data.ts
@@ -198,7 +198,7 @@ const data: Protocol[] = [
     treasury: "synthetix.js",
     twitter: "synthetix_io",
     audit_links: ["https://docs.synthetix.io/contracts/audits/"],
-    oracles: ["Chainlink"],
+    oracles: ["Chainlink","Pyth"],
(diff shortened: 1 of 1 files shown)


mariobern:
No numbers unfortunately but we're pretty much SNX primary oracle on Optimism 

We had this discussion a few months back about being able to add TVS for a protocol in a specific chain (provided the protocol is on multiple chains)

0xngmi:
so atm the way our code works is that all tvl of a protocol is attributed to their oracles, we don't have a way to split it, doing so would require new changes to our backend code

mariobern:
Would it be possible to add Pyth alongside Chainlink as the oracle for Synthetix for now? and once we can split the TVL, we don't mind to be placed to be just under Optimism

mariobern:
https://dune.com/queries/1898719/3129122

Based on this Dune stats, TVL for SNX on Optimism is around 230M

PR #2697, opened 2023-03-29
postgres

waynebruce0x:


coins/src/triggerFetchCoingeckoData.ts
```diff
@@ -1,13 +1,14 @@
+require("dotenv").config();
```
0xngmi, line 1: will dotenv work here on backend?

PR #2702, opened 2023-03-29
Rename Arbitrum to Arbitrum One & Update Logos

bfreshhb:
Hey frens, just made a handful of changes

Mainly around renaming "Arbitrum" to "Arbitrum One". Feel free to double check to make sure I didn't break anything lol, tried to make sure the changes were consistent across the board.

Also updated the logos again since the ones I sent earlier don't look too good in the that format.


0xngmi:
need to make some other changes to this since all adapters are using the `arbitrum` key (no need to change adapters, we just need to handle that). However since everybody knows arbitrum one as simply arbitrum i'm not sure if changing the name would just confuse people more.

Previously we didnt rename binance chain because most ppl know it by "BSC"

PR #3020, opened 2023-04-19
Add Aktionariat 

rube-de:
Add price adapter for security tokens traded over the Aktionariat platform. 
Confidence hard coded to 1 as this is the only official price source for these security tokens.


luziusmeisser:
You are right, I'll check with the team.

luziusmeisser:
The team has confirmed that the responsible graph node is offline. We plan to revise the api and provide an update.

luziusmeisser:
We might do a larger change in a few weeks. But for the time being, we should stay with the current subgraph.

PR #3064, opened 2023-04-22
update merlin info

CryptoSnowPrince:
I updated merlin project infomation


Define101:
Why remove the fork field information? 
Also, we don't use the field audit_notes, so if you could remove it

CryptoSnowPrince:
That's because this project is not same logic to uniswap.
And I agree to remove the field audit_notes if llama team don't use.

CryptoSnowPrince:
I removed audit_note

CryptoSnowPrince:
Thank you for your merge.
I want to ask why I can't see only 2 tokens on tvl charts.
As you kwow, in my project, there are some other tokens such as MAGE, MUTE etcs.

PR #3477, opened 2023-06-14
Feat: add oracle for Ultronswap (Witnet)

parodyBit:
[https://ultron.foundation/blog/ultron-x-witnet-a-powerhouse-partnership-for-cryptoeconomics](https://ultron.foundation/blog/ultron-x-witnet-a-powerhouse-partnership-for-cryptoeconomics)
[https://medium.com/witnet/witnet-deploying-on-ultron-network-bcebccaacd78](https://medium.com/witnet/witnet-deploying-on-ultron-network-bcebccaacd78)
[https://feeds.witnet.io/ultron](https://feeds.witnet.io/ultron)

ultron_proof_of_oracle

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -13716,7 +13716,7 @@ const data2: Protocol[] = [
     cmcId: "21524",
     category: "Dexes",
     chains: ["Ultron"],
-    oracles: [],
+    oracles: ["Witnet"],
     forkedFrom: ["Uniswap V2"],
     module: "ultronSwap/index.js",
     twitter: "ultron_found",


Define101:
In case Witnet oracle fails, are funds at risk in Ultronswap?

parodyBit:
> In case Witnet oracle fails, are funds at risk in Ultronswap?

it is important to consider that a query can fail for different reasons.
at the application level-> the different sources a dApp relies on could return an invalid http response (404 etc)
at the witnet level-> the required consensus for the transaction might not be reached, or there might not be enough interested  witnessing nodes due to low incentives ("Witnet: Tally: Insufficient commits")

if either of those two instances occur -> the dApp relying on the price feeds would use the last posted response in the price feed contract storage-  which could be an old/stale value. 

it is also important to consider the oracle design of witnet on how data is actually updated and verified.

realdealshaman:
@parodyBit what is the oracle used for in Ultronswap? we only track the tvl of the liquidity pools there

PR #3548, opened 2023-06-26
Remove Doveswap v2 and change logo

dovishfinance:
Hello, we'd like to remove DoveSwap V2 as it is no longer supported—we have completely migrated to V3 on our protocol. We've also updated our logo. Here are the changes:

- Remove DoveSwap V2 and make DoveSwap standalone (no parent).
- Change the formatting of the protocol's name: "DoveSwap" instead of "Dove Swap"
- Update DoveSwap's logo

Thank you!


0xngmi:
we'd rather keep old protocols because:
- they still contribute to the historical tvl of their chain
- they have valuable historical data

dovishfinance:
@0xngmi Gotcha, that makes sense. The reason why we wanted to not have v2 was because the volume of our protocol gets hidden because of missing volume data for v2. Is there a way for us to show our volume here? Perhaps we can make our V3 listing standalone and not siblings with v2?

dovishfinance:
@0xngmi alternatively, is there a way to set our volume for our older (v2) protocol to 0 so the parent protocol can show the volume?

PR #3605, opened 2023-07-03
Vaultka: add tvl info for new strategies sake and whiskey

0xKMG:
Update the TVL info for new Strategies of the Protocol, sake and whiskey. Both of them have a leverage vault and lending vault, borrowed assets in lending vault are deducted from the TVL calculation


0xngmi:
if you want to change TVL you need to make a PR on https://github.com/DefiLlama/DefiLlama-Adapters

Your current PR seems to be copying code from one file used for refilling user data into another related file

PR #3631, opened 2023-07-05
Predy finance, update audit links

wshino:
I'm one of the developers in Predy finance. We have used Immunefi.


realdealshaman:
Hi, @wshino has Immunefi done an audit? We'd rather add the link to an audit if one is available

wshino:
@realdealshaman Thank you for reviewing. For v3.2 the audit has not been finished yet, let me remove audit link, and let me add audit link for v2.

PR #3724, opened 2023-07-15
Remove Oracle from Pangolin

sleezyggg:
TVS methodology by @0xngmi here: https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254

Pangolin is a DEX that issues LP tokens. Redstone prices these LP tokens to be used by other protocols to utilize in e.g. lending platforms. You can read about that [here](https://blog.redstone.finance/2023/05/19/redstone-x-pangolin-unleashing-defi-potential/).

It is important to note that Pangolin's TVL is in no way secured by Redstone. They merely price their LP tokens to be used by other protocols. One such integration is [DeltaPrime](https://defillama.com/protocol/deltaprime), that allows Pangolin LP tokens to be used on Avax. The TVL of DeltaPrime is hence correctly attributed to Redstone's TVS. The Pangolin TVL is wrongly attributed to Redstone's TVS, since it is a) not securing any value on Pangolin directly and b) double counted here as the value securing occurs on DeltaPrime. 

This PR is created to enforce the Defillama TVS methodology and remove TVS attributed to Redstone from Pangolin.

PS: Note also that the Redstone announcement specifically mentions an integration on AVAX only, yet it was attempted to attribute all Pangolin TVL across all chains to their TVS. This is in addition to not actually securing any TVL on Pangolin directly and apparently a trend among a lot of oracle projects to incorrectly attribute TVS to themselves.

### DIFF
--- a/defi/src/protocols/data.ts
+++ b/defi/src/protocols/data.ts
@@ -2968,7 +2968,6 @@ const data: Protocol[] = [
     module: "pangolin/index.js",
     twitter: "pangolindex",
     forkedFrom: ["Uniswap V2"],
-    oracles: ["RedStone"],
     governanceID: ["snapshot:pangolindex.eth"]
   },
   {


mattgurbiel:
hey @sleezyggg 
cc @realdealshaman 

Here is the Rationale we have included in our original PR requesting to add Pangolin to RedStone: 
_A potential misprice would mean a cascade of liquidations and automatic withdrawal of Pangolin's TVL.
RedStone <> Pangolin codebase: https://github.com/redstone-finance/redstone-oracles-monorepo/blob/main/packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/dex-lp-tokens/PangolinLP.abi.json ._

According to the methodology outlined here:
https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254
_If there's a case in the future where this will actually impact final TVS numbers meaningfully we'll change our backend to properly reflect the TVS secured by each oracle._

DeltaPrime positions (https://app.deltaprime.io/#/prime-account/assets) opened on Pangolin now constitute roughly $8M (this changes over time) so a potential misprice would mean the loss of around ~35% of Pangolin's TVL. I believe this is meaningful damage to the protocol - perhaps a change in the DefiLlama backend could be needed to calculate that properly.

mattgurbiel:
On another note - LP tokens are a new theme in defi and redstone is one of the first oracles to provide that. It is therefore difficult to find a 100% accurate analogy with other popular models such as lending.

sleezyggg:
> hey @sleezyggg cc @realdealshaman
> 
> Here is the Rationale we have included in our original PR requesting to add Pangolin to RedStone: _A potential misprice would mean a cascade of liquidations and automatic withdrawal of Pangolin's TVL. RedStone <> Pangolin codebase: https://github.com/redstone-finance/redstone-oracles-monorepo/blob/main/packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/dex-lp-tokens/PangolinLP.abi.json ._
> 
> According to the methodology outlined here: [DefiLlama/DefiLlama-Adapters#6254](https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254) _If there's a case in the future where this will actually impact final TVS numbers meaningfully we'll change our backend to properly reflect the TVS secured by each oracle._
> 
> DeltaPrime positions (https://app.deltaprime.io/#/prime-account/assets) opened on Pangolin now constitute roughly $8M (this changes over time) so a potential misprice would mean the loss of around ~35% of Pangolin's TVL. I believe this is meaningful damage to the protocol - perhaps a change in the DefiLlama backend could be needed to calculate that properly.

But this process is started because Pangolin LP positions are being used as collateral on DeltaPrime. 
The fact that you're using an LP token as collateral and that the underlying DEX loses TVL because another protocol that accepts that LP as collateral and is forced to unwind the LP because of a liquidation, does not constitute to TVL that is attributed to the DEX being attributed to the Oracle TVS.  

Uniswap TVL also was never counted towards MakerDAO oracles just because MakerDAO allowed you to mint DAI with Uni LP positions, which would cause drop in TVL of Uniswap when liquidated. It is a flawed analogy. 
Using LP as collateral is nothing new. Counting DEX TVL towards an Oracle because a lending market is allowing that LP to be used as collateral and a subsequent liquidation will lead to a loss in TVL **is new**. 

Not only has this never been done (as in couting LP tokens of a DEX towards TVS because it's lendable/borrowable), but you're counting the same thing (the LP) as TVS on both DeltaPrime and the underlying DEX.

PR #3925, opened 2023-08-02
betswirl new token

Define101:


RomualdH:
@Define101 We are also deployed on Avalanche and Arbitrum.

PR #4001, opened 2023-08-09
HashMix Protocol: move into Liquid Staking

kylidboy:
Our liquidity token is cWFIL, which is deployed on https://filfox.info/en/address/0x587A7eaE9b461ad724391Aa7195210e0547eD11d. After user stake on HashMix protocol, he will receive certain amount of cWFIL which represents his stake, which can be sold, transferred or involved in any other DeFi operations.


Define101:
Sorry, is that the token address for cWFIL
Also where can you sell or use in other defi ?

PR #4056, opened 2023-08-15
add covo finance oracle

mariobern:
gm llama sirs, 

requesting your review to add Pyth as the oracle for both Covo V1 and V2

https://covo-finance.gitbook.io/covo/trading/orders/order-settlement

https://covo-finance.gitbook.io/covo/trading/orders/order-settlement

thank you

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -25271,7 +25271,7 @@ const data2: Protocol[] = [
     chains: ["Polygon"],
     module: "covo-finance/index.js",
     twitter: "covofinance",
-    oracles: ["Chainlink"],
+    oracles: ["Chainlink","Pyth"],
     forkedFrom: ["GMX V1"],
     parentProtocol: "parent#covo-finance",
     listedAt: 1676226573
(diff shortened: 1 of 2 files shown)


realdealshaman:
Hi @mariobern , thank you for the PR, can you provide info on the markets that use Pyth? It is my understanding from the info in the link that only markets not supported by chainlink use Pyth

mariobern:
Hey, yes so I believe the way it works is that currently all markets use Pyth, with Chainlink as the secondary oracle. If Covo lists a market that's not supported by Chainlink, it will use only Pyth as the oracle.

I can get you a written confirmation from the Covo team that this is indeed the case

realdealshaman:
@mariobern that would be great to have here for reference, I was looking for a discord for them but could not find it. We can add Pyth as the only oracle if it is the main oracle 👍

mariobern:
Covo
@realdealshaman let me know if you need anything else sir, thank you 🙏

0xngmi:
could you ask them what would happen if pyth price fees were wrong? could all tvl be drained?

mariobern:
@0xngmi @realdealshaman attaching the answer from Covo

Covo pt2

PR #4103, opened 2023-08-19
satori change name

miler012:
change the name form "Satori Finance" to Satori;
because our raise submit use the name Satori not show at the dashbord.


Define101:
https://github.com/DefiLlama/defillama-server/compare/master...Define101:defillama-server:master
Should be live within 1 hour

PR #4140, opened 2023-08-23
update categories

lgingerich:
Add "Rollup" category to Base and zkSync Era


realdealshaman:
Hey, thanks for the PR, these were previously added. Thanks again

lgingerich:
It's still showing up incorrectly on the Fees dashboard. I made this change to fix that. 

Only on mobile rn so maybe I missed another change that just got made. image

PR #4232, opened 2023-09-03
zkswap finance update token infomation and token price

ZkSwapFinance:


ZkSwapFinance:
Hello @Define101, Please tell me is there any problem with our PR? We want to update the token information but it's been 2 days and we haven't received a response.
thanks,

PR #4365, opened 2023-09-14
add new protocols to RedStone

Kazik435:
gm,

requesting to add RedStone as an oracle for Gravita, Raft and Angle. Submitting for your review:

1. Gravita:
-> https://docs.gravitaprotocol.com/gravita-docs/how-does-gravita-work/vessels-and-collateral#7.-oracles
-> https://twitter.com/redstone_defi/status/1701913677031727333

2. Raft:
-> https://docs.raft.fi/raft-protocol/how-is-r-backed/cdp-vaults/active-vaults/interest-rate-vaults-new/sweth-vault
-> https://twitter.com/redstone_defi/status/1702245848841965989

3. Angle:
-> https://twitter.com/redstone_defi/status/1689652926112047104
-> RedStone is an accepted Oracle in Angle's docs and below also contract with usage
https://docs.angle.money/transmuter/mintburn#target-price-and-deviation

A/ Angle's contract that uses RedStone https://etherscan.io/address/0x00253582b2a3FE112feEC532221d9708c64cEFAb#readProxyContract
B/ You can read as proxy: getOracle(0x2F123cF3F37CE3328CC9B5b8415f9EC5109b45e7), where the address is the C3M one
C/ get the oracleData field, now you need to decode it like that https://github.com/AngleProtocol/angle-transmuter/blob/6e1f2eb1f961d6c3b1cdaefe068d967c33c41936/contracts/transmuter/libraries/LibOracle.sol#L119. 
Or you can just sport Redstone C3M oracle —> 0x6E27A25999B3C665E44D903B2139F5a4Be2B6C26

Thanks,
Marcin

PR #4464, opened 2023-09-27
Adding RedStone oracles to Raft

mattgurbiel:
gm
we’d like to revisit adding Redstone as an Oracle for RAFT protocol.

**Materials**
-> https://docs.raft.fi/raft-protocol/how-is-r-backed/cdp-vaults/active-vaults/interest-rate-vaults-new/sweth-vault
-> https://twitter.com/redstone_defi/status/1702245848841965989

**Rationale**
We agree that 2 weeks ago it was too early to add it: 
https://github.com/DefiLlama/defillama-server/pull/4365 
and @Define101 comment: _The only pool that use Redstone oracle as primary oracle have a very low TVL (around 150k)_ 

However, in the meantime, there have been two updates:
- The total supply of swETH grown from ~$150k to ~$413k (deposited on the https://etherscan.io/address/0x9ab6b21cdf116f611110b048987e58894786c244#readContract)
- More importantly, the total debt cap of R tokens that could be minted using swETH is now $3,300,000: https://docs.raft.fi/raft-protocol/how-is-r-backed/cdp-vaults/active-vaults/interest-rate-vaults-new/sweth-vault

If RedStone Oracle delivered a corrupted price of swETH, anyone would be able to mint more than 3M of R tokens (15% of tokens in circulation) from thin air.  It’s worth noting that although the vaults are isolated, the R tokens minted from different vaults are fully fungible. This exploit would drag the R price down at least by 15% (a huge price movement for a stablecoin), but probably even more, taking into account the psychological effect of the unbacked tokens on the market. 

Cheers,
Matt


0xngmi:
raft TVL is 28m but the pool covered by your oracle is only ~417k in TVL, our rules for adding oracles to a protocol where they're only used partially are:
- we'll add it if the oracle-secured value can be separated by chain (eg 1m in ethereum secured by X and 2m in bsc secured by Y)
- we'll add it if TVL secured by an oracle is >50% of total TVL

in this case none of the 2 rules applies here, so we'll have to reject this update

mijovic:
Hey @0xngmi,Đorđe here CTO at Raft.
In our model, pools are not segregated, so we rely on the security of both RedStone and Chainlink. If swETH is priced incorrectly, users could mint “fakely backed’ R tokens, which is unacceptable. Therefore I believe the argumentation of keeping Raft with RedStone and Chainlink as Oracles holds.

PR #4477, opened 2023-09-28
add pyth oracle

mariobern:
gm llama sirs,

requesting your review to add protocols that have integrated Pyth:

- DIP Exchange
- Fuzio 
https://x.com/PythNetwork/status/1691465630808174592?s=20

- Kinetix Derivatives
- Mendi Finance
- SeaSwap

thank you


mariobern:
gm @realdealshaman 

makes sense on the dexes sir

ill remove seaswap and fuzio for this PR

regarding your question on DIP, below is a screenshot of the doc page. if this is not sufficient we'll work with DIP directly to do an announcement on how they use Pyth.

image

PR #4644, opened 2023-10-17
add pyth protocols

mariobern:
gm llama sirs

requesting your review to add 2 new protocols that use Pyth

Angle
https://x.com/PythNetwork/status/1713963107025076433?s=20

Thales
https://x.com/PythNetwork/status/1694002716325126649?s=20

thank you

### DIFF
--- a/defi/src/protocols/data.ts
+++ b/defi/src/protocols/data.ts
@@ -9254,7 +9254,7 @@ The eWIT token is a custodial, wrapped version of the Witnet coin managed by the
     chains: ["Ethereum"],
     module: "thales/index.js",
     twitter: "thalesmarket",
-    oracles: ["Pyth"],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @mariobern Thanks for the PR, While Pyth oracles are being used in both Angle and Thales, we need to be cautious about how we represent the total value secured by Pyth. Since the value secured by Pyth within Thales is relatively low it wouldn't be accurate to add the entire TVL under its TVS.

A potential compromise could be to add only the TVL of Polygon on Thales under the TVS of Pyth, since the speed markets there account for about 50% of the TVL. But, on other chains, it wouldn't provide a fully accurate representation

mariobern:
gm ser @realdealshaman, thank you for the comment

removed Thales for now

PR #4754, opened 2023-10-25
Update Umbrella Network data

ed-umb:
Update project URL, chains in which the project is available, audit status and audit_links.


realdealshaman:
@ed-umb  we can update the audits np but, the "chains" portion of the listing corresponds to the chains where we are tracking tvl for umbrella in the tvl adapter: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/umbrella/index.js

To add extra chains we would need to first update the tvl code

Also, we prefer to use the staking url since it goes with what we are tracking as tvl, is the url for staking UMB correct?

ed-umb:
Hello @realdealshaman,

As per the documentation at the beginning of the data.ts, field 'chains' is
not being used in the current code. Should I still remove those blockchains
for which there is not a fetcher?

Regarding the URL, yes, we can keep the staking url. Should I cancel this
PR and open a new one?

Regads

realdealshaman:
@ed-umb yeah, adding the chains without having them in the tvl adapter will cause a error when trying to deploy the changes on our server, I've updated the file so that we keep the audit links

ed-umb:
Thank you
*_______________________*
Eugenio Di Santo
telegram: @edisanto
www.umb.network

PR #4854, opened 2023-11-06
Update data.ts

BeefyPower:
Updated Beefy.com information following token migration from BSC to Ethereum


Define101:
Thanks for the PR
We have made some changes here https://github.com/DefiLlama/defillama-server/pull/4856
The "chain" field should not be changed, It represents the first blockchain where that protocol launch (even if they are not there anymore)

PR #4900, opened 2023-11-10
fix(abracadabra): Update chains and oracles

0xmDreamy:


realdealshaman:
@0xmDreamy our tvl adapter does not track the new chains you've added to the listing: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/abracadabra/api.js

We would need to have a value exported for them first before adding them to the data.ts file

PR #4902, opened 2023-11-11
Sommelier: Update Oracles

philipjames44:
Requesting to add Chainlink and RedStone Oracles to Sommelier

Materials:
https://sommelier-finance.gitbook.io/sommelier-documentation/smart-contracts/oracle-usage-and-asset-pricing
https://twitter.com/sommfinance/status/1692567027787923532

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -17016,7 +17016,7 @@ const data2: Protocol[] = [
     cmcId: "18248",
     category: "Yield Aggregator",
     chains: ["Ethereum"],
-    oracles: [],
+    oracles: ["Chainlink", "RedStone"],
     forkedFrom: [],
     module: "sommelier/index.js",
     twitter: "sommfinance",


sleezyggg:
gm @realdealshaman. 

As per [this](https://community.sommelier.finance/t/sips-071-upcoming-turbo-sweth-cellar-proposal/1170/2) governance proposal Redstone is utilized for the swETH Cellar only. 

The swETH Cellar currently has 4.7M in TVL. 

All other cellars are utilizing Chainlink Oracles and account for the remaining 57.3M in TVL.
Is it justifiable to attribute over 62M in TVS to Redstone when they are securing a single cellar that represents 7.5% of protocol TVL, especially given that if anything happens to that cellar the remaining 57.3M remain untouched?

image

philipjames44:
As far as I know there's no way to split up oracle TVL fractionally for protocols, if I'm mistaken please let me know. 

Regarding Redstone, you're correct that they currently account for a subset of Sommelier's total TVL, however they're a dedicated team that's been a pleasure to work with, and we will continue to add their oracles in upcoming vaults as is appropriate.

Barring for the ability to split up TVL fractionally, I'd like to vouch for them to continue to be included, as they are indeed an active & valued oracle provider for us.

sleezyggg:
> As far as I know there's no way to split up oracle TVL fractionally for protocols, if I'm mistaken please let me know.
> 
> Regarding Redstone, you're correct that they currently account for a subset of Sommelier's total TVL, however they're a dedicated team that's been a pleasure to work with, and we will continue to add their oracles in upcoming vaults as is appropriate.
> 
> Barring for the ability to split up TVL fractionally, I'd like to vouch for them to continue to be included, as they are indeed an active & valued oracle provider for us.

As per [this](https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254) it should be removed until the covered market rises significantly:

>If an oracle only covers a small part of a protocol and failure of it would only mean a small loss for the protocol, for example a protocol that uses oracle X for almost all their pools but there's a pool that accounts for 5% of TVL that uses oracle Y, such that a failure of X would lead to 95% TVL loss and a failure of Y would lead to 5% TVL loss. In this case we simply count all the TVL towards X's TVS and ignore Y. This is slightly inaccurate, but these cases are extremely rare and in aggregate the amount of TVL that falls under these is tiny compared to total TVS numbers, so it's negligible (think <0.1% change). If there's a case in the future where this will actually impact final TVS numbers meaningfully we'll change our backend to properly reflect the TVS secured by each oracle. 

@0xngmi also agrees [here](https://discord.com/channels/823822164956151810/823822533974425621/1184658672111058955).

PR #4917, opened 2023-11-14
feat: update diamond protocol info

Chi-AnTai:
We want to rename our product from UnibotV2 to Diamond Protocol V2


Define101:
Thanks for your PR
We made some adjusts here https://github.com/DefiLlama/defillama-server/pull/4940

PR #5030, opened 2023-11-24
add bluefin oracle

mariobern:
gm llama sirs

requesting your review to add the oracles for Bluefin

https://learn.bluefin.io/docs/bluefin-101/oracle-and-index-price

thank you very much


Define101:
Please use the oracleByChain instead oracle field please
image

PR #5060, opened 2023-11-28
Modify Category - Ondo Finance

tom2o17:
According to the category descriptions linked [here](https://defillama.com/categories) 

We feel as though Ondo better falls into the RWA-Lending category as it tokenizes real world assets for use as collateral. 
As opposed to the tokenization of physical non-traditional-financial assets.

Ty 🫡

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -26248,7 +26248,7 @@ const data2: Protocol[] = [
     audit_note: null,
     gecko_id: "ondo-finance",
     cmcId: null,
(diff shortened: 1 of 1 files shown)


realdealshaman:
@tom2o17 we would need to be able to distinguish between tvl and borrowed funds so that we can subtract borrowed from the tvl and export borrowed funds separately under the borrowed label to categorize it as RWA Lending

realdealshaman:
closing due to inactivity, please request to re-open if more info can be provided. thanks

PR #5077, opened 2023-11-30
Update Amulet Protocol Product Name

yym-MetaRisk:


realdealshaman:
@yym-MetaRisk is it not the V2?

yym-MetaRisk:
> @yym-MetaRisk is it not the V2?

@realdealshaman Thanks for reply. Its V2. To distinguish between V1 and V2, we want to use Amulet V1 for V1 and Amulet Protocol for V2.

alexandrosgoh:
@realdealshaman we have updated some of the description and revert back the name changes to V2

PR #5079, opened 2023-11-30
Add wan-bridge cross-token price mapping.

lolieatapple:
Add wan-bridge cross-token price mapping.


waynebruce0x:
I have reverted this PR because some of the decimal values were incorrect and caused issues with our data. If you'd like to discuss further, please message our discord since we dont monitor comments on closed PRs, thanks!

PR #5106, opened 2023-12-04
Change Orchai Protocol Info

trung2891:
change:
- category: "Liquid Staking"  --> "Services"
- chains: add "Binance"
- forkedFrom: remove


realdealshaman:
@trung2891 "Services" is to broad of a category, we prefer to be more specific where possible, the product we are tracking, while considering the tvl code is "Liquid Staking" correct? Can we close this PR?

PR #5206, opened 2023-12-13
fixed forkedFrom

ZkSwapFinance:
Reason: more and more contracts forked from Uniswap v2 & Pancakeswap than Syncswap. Additionally, Syncswap is also forked from Uniswap.


Define101:
Hello, you are completely correct. This should always had been marked as uniswap v2 fork
Thanks for pointing this, and sorry for this situation
Merged

PR #5208, opened 2023-12-13
add api3 user

umersin61:
Hey Llamas. 

Metavault Perps v2 is utilizing a double oracle approach with pulling prices from Pyth and checking them against API3. 
Traders hence can only open a position if Pyth & API3 are within a certain range from each other.

https://x.com/MetavaultTRADE/status/1734226620591043063?s=20

Thank you!


umersin61:
little more context here: https://docs.metavault.trade/more/smart-contract-overview/v1-and-v2-perps

Bildschirmfoto 2023-12-13 um 10 49 05

PR #5233, opened 2023-12-15
add solend to Pyth

mariobern:
gm 

requesting your review to add Pyth back to Solend

not sure why Solend was taken out from Pyth as it's the primary oracle for the main pools

thank you


mariobern:
https://docs.solend.fi/permissionless-pools/pre-listing-checklist#oracles

image

additional details attached above

PR #5280, opened 2023-12-20
remove "Kamino Liquidity" from Pyth TVS

sleezyggg:
This is a PR to remove the Liquidity Product of the Project Kamino from Pyth TVS. 

[According to their documentation,](https://docs.kamino.finance/products/liquidity) using Kamino Liquidtiy, you can pool assets and receive kTokens for them that are in effect simply LP tokens (no oracle involvement, also not found in their documentation). 

These tokens are then useable in their lending product [Kamino Lend](https://defillama.com/protocol/kamino-lend). 
Not only, is the TVS being double counted here, as Kamino Liquidity kTokens are all TVL for Kamino Liquidity and effectively nearly all TVL in Kamino Lend (because the kTokens are simply deposited there), but i couldn't find any evidence for Kamino Liquidity actually utilizing oracles for the creation of these kTokens.

The only thing that utilizes oracles is Kamino Lend (to price these LP tokens) and as such Pyth should be kept as an oracle on Kamino Lend, but removed as an oracle on Kamino Liquidity.

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -15299,7 +15299,7 @@ const data2: Protocol[] = [
     cmcId: null,
     category: "Liquidity manager",
     chains: ["Solana"],
-    oracles: ["Pyth"],
+    oracles: [],
     forkedFrom: [],
     module: "kamino/index.js",
     twitter: "Kamino_Finance",


sleezyggg:
In fact, the only mentioning of oracles is in the K-Lend section [here](https://docs.kamino.finance/kamino-lend-litepaper#oracles), where Pyth and Switchboard are mentioned. As such i would also request Switchboard to be added to Kamino Lend. 

This still means Pyth should be removed from Kamino Liquidity.

mariobern:
Hey all,

seems that this PR was merged before even clarifying to the Pyth team or the Kamino team. We are very disappointed with the lack of communication that has transpired in multiple cases involving Pyth over the last few weeks.

We have been working with Hubble and Kamino team since the start of their launch in 2021 until now, continually supporting new assets and vaults.

Here's a screenshot where we clarified to the Kamino team regarding Pyth as the oracle for Kamino Liquidity.


image

 In future cases, we would very much appreciate it if we are given the heads up, or if any future PRs from a third party could get verified by core contributors of the protocol.

sleezyggg:
gm @0xngmi and @mariobern.
agree that this was merged too quickly and that the projects should be given time to produce evidence, like i asked for in discord. 

@mariobern could you ask kamino to add this to their docs for their liquidity product so that independent third-parties like me can verify such things and don't need to rely on some telegram dm screenshots?
As it stands right now there is no evidence in their docs available and we only have your screenshot as "proof". 
Thanks!

0xngmi:
the reason why i merged it here was that i just checked it by myself, on their app the top vaults by tvl are all stable pairs, driving more than >50% of tvl
Screenshot from 2023-12-21 11-49-57

there's no reason for rebalances to happen on these pairs and if there is one it doesnt seem like oracles are needed, plus i checked kamino's docs and they say that usdt-usdc has never rebalanced and same for other stable pairs

that's why i merged the PR, in case I'm wrong I'll apologize and revert the merge

0xngmi:
I've reverted the change in 00d57a80f95fae276e890ad3cfd354fe0137a998 while we wait for Kamino's input

0xngmi:
from kamino's team
> The SOL-XSOL strategies are deterministic drift strategies. The range at which we provide liquidity is unaffected by the market price or oracle prices

so it seems like all the SOL pools that account for most of the TVL do not use oracles

a-person12:
@0xngmi you didn't ask,

 "does Kamino use oracles?" 

you asked: 

"if pyth were to be hacked and report incorrect prices, would a hacker be able to drain >50% of kamino liquidity's TVL?" 

would it be better if you got an explanation exactly how and where Kamino uses oracles?

a-person12:
your question: 

"Given that >50% of TVL in Kamino Liquidity is in correlated pairs (eg sol-jitosol), if pyth were to be hacked and report incorrect prices, would a hacker be able to drain >50% of kamino liquidity's TVL?"

this was v specific but seems to have lead to a conclusion that kamino does not use oracles 

we have 3 diff places of async comms between 3 projects, this doesn't seem ideal. Happy to get us all properly on the same page

oeble:
Hello @0xngmi , I'm Olivier (aka. CaptainRusty) from the Kamino team. We do use Pyth, for the following assets for both Liquidity and Lending products: SOL, ETH, BTC, MSOL, USDC, USDT, HNT, RENDER, bSOL, PYTH, JTO, JITOSOL, LST and more to come.

Even in the case of deterministic drift strategies, we don't need oracles only for the range choice, but for other things, for example, the oracles are still used as a reference value to determine if a swap from one token to another during a rebalance is done at a fair price. We also use the Oracle price to compute the USD value of deposited/invested tokens itself used to compute the price of an LP share (kTokens) and the number of shares issued to a new depositor depending on the setup of the strategy. We also use the EMA price of Pyth to determine if there is any anomaly with the prices and to ensure we don't rebalance when prices are unreliable.

I hope this clarifies our usage.

0xngmi:
The most important question is
> if pyth were to be hacked and report incorrect prices, would a hacker be able to drain >50% of kamino liquidity's TVL?

Since that's what we use in our methodology to determine if we should include the TVL of a protocol under an oracle TVS or not, that is, if the oracle is securing that protocols TVL.

Could the usages that you mention lead to a loss of >50% of TVL if the oracle was incorrect and reported incorrect data?

oeble:
It is part of the protocol security, and manipulation of the oracles can affect all the TVL. But if it is an oracle manipulation alone, then it should not lead to a loss of > 50% of TVL because we have other security mechanisms to prevent this level of exploit from happening.

To be fair, if this is your definition of TVS, I personally would not trust any of the protocols listed as "secured by an oracle" in DefiLlama's TVS list.

TLDR;

To this:

> if pyth were to be hacked and report incorrect prices, would a hacker be able to drain >50% of kamino liquidity's TVL?

My answer is no.

To this:
> Is Pyth used to protect more than 50% of Kamino liquidity TVL.

My answer is yes.

PR #5293, opened 2023-12-20
Change Phoenix Trade to Phoenix

0xkrane:
protocol's name is phoenix and not phoenix trade

also it seems like the update broke volume for phoenix being displayed on the defillama website? is this just a transient break or should we look for what happened


Define101:
Checking this

Define101:
https://github.com/DefiLlama/defillama-server/pull/5296

PR #5327, opened 2023-12-26
Add oracle data for Thales to data.ts

scvija:


scvija:
Hi there @Define101, i have seen in the PR that the guys from WINR opened you were asking around how does Pyth secure their contracts, so i wanted to post some links preemptively.
Here is the Speed Markets contract using Pyth:
https://github.com/thales-markets/contracts/blob/main/contracts/SpeedMarkets/SpeedMarketsAMM.sol
Here is a link to the Thales docs explaining the usage:
https://docs.thalesmarket.io/thales-parimutuel-markets/speed-markets#how-speed-markets-work

Here are some examples for Chainlink also:
https://github.com/thales-markets/contracts/blob/8c9193606a2930df1a06ffb42e0475bc61b6f20f/contracts/SportMarkets/Rundown/TherundownConsumerWrapper.sol#L6
Docs:
https://docs.thalesmarket.io/using-thales/thales-amm-design#thales-amm-mechanism
https://docs.overtimemarkets.xyz/sports-amm-smart-contract

Define101:
Hello @scvija thanks a lot for the explanation.
Regarding speed Markets products, seems that its only a very small amount of your total TVL correct? Less than 1% of the tvl

Define101:
Regarding chainlink looks good
Just one thing, we are able to divide the oracles per chain, so let us know, if x oracle only works on y chain (since you guys are in quite a few chains)

scvija:
> Hello @scvija thanks a lot for the explanation. Regarding speed Markets products, seems that its only a very small amount of your total TVL correct? Less than 1% of the tvl

Hi there @Define101 , that is correct. The liquidity of speed markets is a very small chunk, since most of the markets created have a very short expiration, so the liquidity is not held up for long periods of time. The largest amount of markets are 10min, 20min or 60 min markets, so the liquidity can be reused and is capital-efficient.

Regarding the second point, we have:
- Digital Options markets (using Chainlink)  on OP, ARB, Polygon and Base,
- Speed Markets(using Pyth) on OP, ARB, Polygon and Base
- Sports Markets (using Chainlink) on OP, ARB and Base
So I would say we have both oracles on all 4 of these networks. 
If you need some adjustments, let me know.
Best regards!

Define101:
since the speedmarkets products is only a very small amout of TVL , we can not add it to the oracles. Our rule is around 50% TVL https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254

Define101:
We made some changes here https://github.com/DefiLlama/defillama-server/pull/5337

PR #5461, opened 2024-01-08
Update tokenMapping for hyUSD (Base)

larrythecucumber321:
This change adds hyUSD (minted [here](https://register.app/#/overview?token=0xCc7FF230365bD730eE4B352cC2492CEdAC49383e&chainId=8453)) to DefiLlama's coin price API

### DIFF
--- a/coins/src/adapters/tokenMapping.json
+++ b/coins/src/adapters/tokenMapping.json
@@ -265,6 +265,11 @@
       "decimals": "18",
       "symbol": "rcbETH",
       "to": "coingecko#ethereum"
+    },
+    "0xcc7ff230365bd730ee4b352cc2492cedac49383e": {
+      "decimals": "18",
+      "symbol": "hyUSD",
+      "to": "coingecko#high-yield-usd-base"
     }
   },
   "aptos": {


coins/src/adapters/tokenMapping.json
```diff
@@ -265,6 +265,11 @@
       "decimals": "18",
       "symbol": "rcbETH",
       "to": "coingecko#ethereum"
+    },
+    "0xcc7ff230365bd730ee4b352cc2492cedac49383e": {
+      "decimals": "18",
+      "symbol": "hyUSD",
+      "to": "coingecko#high-yield-usd-base"
```
Define101, line 272: this is from coingecko terminal, we cant price from that
need to be from coingecko it self

larrythecucumber321: This asset is also on Coingecko itself https://www.coingecko.com/en/coins/high-yield-usd-base

If this approach isn't correct for getting pricing for this asset, could you please point me to the proper way?

larrythecucumber321:
@Define101 sorry, should I be expecting to see the hyUSD price show up here? Or what else needs to be done?

https://coins.llama.fi/prices/current/base:0xCc7FF230365bD730eE4B352cC2492CEdAC49383e

PR #5482, opened 2024-01-10
chore: Dyson Finance update oracles

anna-pelith:
This is an update to use Pyth as the oracle for Dyson Finance.


realdealshaman:
we don't add oracles to our dex listings since we are only tracking the LPs in the AMM, no oracle needed to secure those pools

PR #5483, opened 2024-01-10
Delete deprecated 'Aperture LM' project.

gnarlycow:
This change removes the deprecated "Aperture LM" project.  The TVL associated with this project is coming from two deprecated Aperture structured product offerings on Terra and Avalanche in the past; both have been deprecated and therefore we would like to remove this listing.  Aperture's current LM (liquidity management) product is non-custodial, meaning that it doesn't take custody of users' tokens or liquidity positions; it helps users easily rebalance and/or reinvest instantly or scheduled to occur when a specified condition is met, so we keep track of volume-based Total Value Enabled metric instead of TVL.


realdealshaman:
@gnarlycow Hi, and thanks for the pr! I understand you'd like to remove the listing due to the deprecation. While we understand the projects status, we prefer to keep historical data for reference. Therefore, we prefer to keep the listing for historical purposes. Could you confirm if the assets on Avalanche that are included in the TVL calculation still exist? If they don't, we can zero them out, similar to what we did with the Terra assets

PR #5531, opened 2024-01-16
fix: update address for kim

agileurbanite:
TVL is not updating on defillama for Kim need to update address to ensure that defillama updates tvl data:
https://defillama.com/chain/Mode

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -26969,7 +26969,7 @@ const data3: Protocol[] = [
   {
     id: "3913",
     name: "KIM Exchange", // previous Kizuna Exchange
-    address: null,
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hi, @agileurbanite is that the address for the native token of Kim or is it something else?

If tvl should be higher, have you guys recently launched a new version with a new factory address?

agileurbanite:
That's the new version of the factory address

realdealshaman:
@agileurbanite that field is where the native token for KIM exchange will go when you have one, I've updated the factory address in the adapter that calculates the tvl here: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/kizuna/index.js

Thanks for the heads up! :+1:  Please don't hesitate to let us know of any new changes. Cheers

realdealshaman:
The new tvl will show up with the next hourly update

PR #5543, opened 2024-01-16
added Supra as oracle provider for TiFi

ksdumont:
Hi,
We would like to add the use of Supra's Oracle to TiFi. TiFi utilize the data feeds from Supra to secure the Crypto Loan asset values.

Materials
-> https://docs.tifi.net/#/TiFiBank/Introduction

Rationale
TiFi utilize the data feeds from Supra to secure the Crypto Loan asset values. If the oracle were to fail, customer's assets would be at risk.

Thank you in advance.

Best,
Keith


realdealshaman:
looks like only INJ and SEI price is being pulled from Supra, it only makes up about $5000 of deposits, needs to be 50% of the tvl to be added

PR #5544, opened 2024-01-16
added supra as oracle provider  to Typus Finance

ksdumont:
Hi,
We would like to add the use of Supra's Oracle to Typus. Typus utilize the data feeds from Supra.

Materials
-> https://docs.typus.finance/options-specifications

Rationale
Typus utilize the data feeds from Supra. If the oracle were to fail, customer's assets would be at risk.

Thank you in advance.

Best,
Keith


realdealshaman:
looks like only INJ and SEI price is being pulled from Supra, these two only make up about $5000 of deposits, needs to be 50% of the tvl to be added. if more assets are being priced using Supra we can add it once Typus updates the docs :+1:

PR #5575, opened 2024-01-18
adding docsend link for audit trail for Bluefin

ibadia:
Added the link to show the audit of bluefin


Define101:
Hello, could you remove this part?
image
The audit must be publicly visible. Users should not have to add an email in order to see the audit

ibadia:
@Define101  We have updated then link, It is accessible by public now.

PR #5607, opened 2024-01-23
return protocols with dummy.js module in protocols api

mintdart:


g1nt0ki (approved):

0xngmi:
added a small change that would have led to a crash otherwise in some cases

PR #5644, opened 2024-01-25
Add API3 Oracle to Init Capital

Ashar2shahid:
Init capital uses API3 feeds for all of their money markets on Mantle : https://init-capital.gitbook.io/init-capital/borrowing/assets-list#stable-assets

ishare-1706145749

In case its not clear, Init is a decentralized lending and borrowing protocol similar to AAVE and Compund but with the ability for other dAPPs to tap into the borrowing and lending side using "Hooks" . As with any lending and borrowing protocol the TVL is secured by Oracles for things like liquidations and API3 is oracle used by init as stated in their docs

For Mintrest there was a typo, Mintrest is not live on manta, it is live on Mantle :)

ishare-1706145979

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -30130,7 +30130,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Lending",
     chains: ["Mantle"],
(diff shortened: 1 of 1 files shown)


0xngmi:
asking their team to verify

PR #5814, opened 2024-02-09
chore: update steadefi info

0x67:
Update Steadefi info to better reflect current conditions

I also want to update the logo but not sure how to do it so I attached here 
Steadefi_Logo_Square


jefflam:
Commenting to get notifications on any responses as part of the team

PR #5845, opened 2024-02-13
update oracles for kinetix derivs v2

umersin61:
Similar to kinetix dervis (v1), kinetix dervis v2 are powered by Pyth & API3. 
Tweet can be found here: https://x.com/KinetixFi/status/1755722584119750867?s=20

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -31385,7 +31385,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Derivatives",
     chains: ["Kava"],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hi @umersin61 thanks for the PR. Can you list which assets use Pyth and which use API3?

umersin61:
> Hi @umersin61 thanks for the PR. Can you list which assets use Pyth and which use API3?

hi. it's both for all assets. it utilizes and optimistic fast price (pyth) that is always checked against API3 before being allowed to be consumed.

PR #5980, opened 2024-02-26
description bouncebit + modifications 5546

Define101:
description bouncebit + modifications 5546


anna-pelith:
After discussions with our Oracle supplier, we have decided not to include the oracle in our DEX product description on the DefiLlama website. Thank you for merging other necessary modifications into the main branch beforehand. However, we still have some essential changes that will be included in a new PR.

PR #6030, opened 2024-02-29
update roup category

butcher-blaviken:
update roup category


Define101:
Hello, was trying to see how roup works. Seems that its a bridge only correct? You can bridge this assets, back and forth. But you can send  roup token for example from mapo and get other token on bitcoin correct?
image

butcher-blaviken:
Yes, you're right

Define101:
Okok. So the category needs to stay bridge

realdealshaman:
Closing this due to inactivity, feel free to open another if any further review is needed

PR #6059, opened 2024-03-02
update fixes world's twitter handle and protocol display name

btspoony:


defi/src/protocols/data3.ts
```diff
@@ -32991,7 +32991,7 @@ const data3: Protocol[] = [
   },
   {
     id: "4179",
-    name: "Fixes World",
+    name: "Fixes 𝔉rc20",
```
Define101, line 32994: can we use another letter and not that one. Not sure if will work

btspoony: Done

PR #6100, opened 2024-03-06
Updated Zircuit metadata

kbak:
Updated the category of Zircuit Staking


realdealshaman:
Hey @kbak thanks for the PR. We reviewed the contract and since it only allows a few functions, we determined that Farm was a more appropriate category. We also followed this same process with Blast for their pre-launch farm product

PR #6125, opened 2024-03-07
Cygnus Finance: update token address & symbol

greatpie:


greatpie:
Gm! Sorry for the misunderstanding. cgUSD is a stablecoin and we haven't issued a governance token yet. So should this field be for governance tokens and not stablecoins?

greatpie:
> cgUSD is not the governance token correct ?

Sorry for the misunderstanding. cgUSD is a stablecoin and we haven't issued a governance token yet. So should this field be for governance tokens and not stablecoins?

Define101:
> > cgUSD is not the governance token correct ?
> 
> Sorry for the misunderstanding. cgUSD is a stablecoin and we haven't issued a governance token yet. So should this field be for governance tokens and not stablecoins?

Correct. The token field is for the governance token only

PR #6130, opened 2024-03-08
update info for SynFuturesV3

ffhu234:
1. SynFuturesV3 use Pyth as price oracle.

2. Implementation details: Pyth price feeder is used as spot price oracle on SynFuturesV3.  The raw spot price is not used directly but undergoes a specific constraint mechanism before being used. An exponential moving average (EMA) method is applied to the raw price to smooth the fluctuation; Mark price is calculated from the smooth spot price, which is used to calculate the unrealized pnl for the account. For dated futures, time weighted average spot price over the last 30 minutes before expiry is used as the final settlement price to calculate pnl for all participants.

3. Documents:
   https://docs.synfutures.com/docs/User-Guide/Security
   https://docs.synfutures.com/docs/faq/general/#market-integrity--stability
   https://www.synfutures.com/v3-whitepaper.pdf


Define101:
@ffhu234 could you share the oracle contract please

realdealshaman:
@ffhu234 there's no mention of pyth in the docs you provided only, chainlink and the twap in the whitepaper, closing this for now since the changes are not needed. V3 is also on arbitrum and the audits 2 simply denotes that there is an audit, you can find the legend for audits at the top of the data3.ts file

PR #6148, opened 2024-03-11
added_usdc_plus_overnight

Kron1749:
Added usdc plus ovenight


Define101:
Sorry, we already had a listing
https://defillama.com/protocol/overnight-usd+

Kron1749:
> Sorry, we already had a listing https://defillama.com/protocol/overnight-usd+

Hello, that's a USDC Plus,not USD Plus, that's 2 different tokens

Define101:
@Kron1749 its already being tracked here https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/overnight-usdPlus/index.js#L9

Kron1749:
> @Kron1749 its already being tracked here https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/overnight-usdPlus/index.js#L9

Yeah,it is,and TVL updated, thank you
But the problem is that USDC should be in here https://defillama.com/protocol/overnight-finance
But there is no USDC+ token

PR #6309, opened 2024-03-20
Added Supra as an oracle to Shoebill

ksdumont:
Hi,
We would like to add the use of Supra's Oracle to Shoebill. Shoebill utilize the data feeds from Supra as their primary oracle.

Materials
-> https://docs.shoebill.finance/risk/oracle

Rationale
Shoebill utilize the data feeds from Supra to secure the User's funds. If the oracle were to fail, customer's assets would be at risk.

Thank you in advance.

Best,
Keith


Define101:
@ksdumont  since redstone is the one securing STONE price, and since STONE is more than 50% of TVL. We can not add it 
https://github.com/DefiLlama/defillama-server/pull/5982

realdealshaman:
Hey @ksdumont thanks for the PR! I've updated Shoebill V2 manually to reflect Supra as the oracle on Manta, I've updated it manually to avoid merge commits, if you can refork the repo or pull changes before submitting a new pr, that would be very helpful, thanks again for the pr

PR #6353, opened 2024-03-25
update oracle data

ShoebillFinance:
we are using both Supra and Redstone push oracle now as blended version.

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -18982,7 +18982,7 @@ const data3: Protocol[] = [
     module: "shoebillFinance-v2/index.js",
     oraclesByChain: {
       wemix: ["Pyth"], // https://github.com/DefiLlama/defillama-server/pull/5434#issuecomment-1889510580
-      manta: ["Supra"], // https://twitter.com/ShoebillFinance/status/1771149332689166521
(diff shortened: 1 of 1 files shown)


defi/src/protocols/data3.ts
```diff
@@ -18982,7 +18982,7 @@ const data3: Protocol[] = [
     module: "shoebillFinance-v2/index.js",
     oraclesByChain: {
       wemix: ["Pyth"], // https://github.com/DefiLlama/defillama-server/pull/5434#issuecomment-1889510580
-      manta: ["Supra"], // https://twitter.com/ShoebillFinance/status/1771149332689166521
+      manta: ["Supra","RedStone"], // https://twitter.com/ShoebillFinance/status/1771149332689166521 https://twitter.com/redstone_defi/status/1763580020751245496
```
Define101, line 18985: Who is securing STONE price? Cant be both

suropark: primary feeds are secured by redstone oracle, and we use supra secondary feeds with max deviation check

ShoebillFinance: primary feeds are secured by redstone oracle, and we use supra secondary feeds with max deviation check

PR #6394, opened 2024-03-27
Update Envelop project

Kostydenis:
Update chains of Envelop project


defi/src/protocols/data3.ts
```diff
@@ -36122,14 +36122,14 @@ const data3: Protocol[] = [
     symbol: "NIFTSY",
     url: "https://envelop.is/",
     description: `Envelop is a collateralized derivatives protocol.`,
-    chain: "Blast",
+    chain: "Ethereum",
```
Define101, line 36125: Please keep "Blast" in this field.

Kostydenis: Ok, I'll change it back

PR #6483, opened 2024-04-02
fix: handle failing erc4626

0xsambugs:
We realized one of the 4626 adapter tokens we supported was failing due to a problem with the underlying vault. This was causing all calls to fail, so we realized we needed to change the approach. The idea is simple though, if a token fails to report any of the properties, we will simply ignore it and return prices for the rest

I wasn't sure if it made sense to log a warning, but I can do it if you would like me to

### DIFF
--- a/coins/src/adapters/utils/erc4626.ts
+++ b/coins/src/adapters/utils/erc4626.ts
@@ -16,37 +16,37 @@ export async function calculate4626Prices(
   timestamp: number,
   tokens: string[],
   hardCodedAssets?: string[],
-): Promise<(Result4626 | null)[]> {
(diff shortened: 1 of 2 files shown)


coins/src/adapters/utils/erc4626.ts
```diff
@@ -16,37 +16,37 @@ export async function calculate4626Prices(
   timestamp: number,
   tokens: string[],
   hardCodedAssets?: string[],
-): Promise<(Result4626 | null)[]> {
+): Promise<Result4626[]> {
```
0xsambugs, line 19: Removed the `null` option, since we weren't returning `null` at all

PR #6693, opened 2024-04-23
Update Brickken Project

xaler5:
Changed category (it was RWA previously but got recently changed to service without us asking for it). Add one more audit links, more supported chains, correct typo


Define101:
@xaler5 hello hello, would you mind telling us a bit more about the token issue? What do they really represent? Also regarding the price of those assets, how is it calculated? (besides the oracle price)

xaler5:
Hello @Define101 we already provided the TVL calculation method into the TVL description but basically:

- BKN is an utility token that is meant to be used within our protocol to run tokenizations/investments and other actions
- The tokenization protocol will issue a token (an ERC20) and an escrow for each tokenization.
- The tokens created through the protocol can then be sold through token offerings through the generated escrow defining a price in USD for each one of those.

The TVL is calculated as the sum of each issued token's total supply within our protocol, each of them multiplied by the price of their latest offering.

So let's say we have tokenized three projects, token A, token B and token C.

The TVL is TVL = (tokenA.totalSupply * tokenALatestPrice) + (tokenB.totalSupply + tokenBLatestPrice) + (tokenC.totalSupply + tokenCLatestPrice)

Define101:
@xaler5  Hello, thanks a lot for your reply
So far what did Brickken tokenize? Real Estate, equities?
Any concrete examples?

xaler5:
@Define101 is there any blocker from merging this ?

PR #6716, opened 2024-04-25
feat: update our dex info details

waliair:
Update details of the HerculesDEX


Define101:
@waliair why remove token address and symbol?

waliair:
@Define101 we officially launched our protocol token, is it wrong syntax?

PR #6800, opened 2024-05-03
Update Grizzly Trade Oracle info. (Pyth supported now.)

0xrmvdao:
Grizzly Trade and Trade V2 oracle info updated.


realdealshaman:
@0xrmvdao thanks for the PR! Can you provide a reference link that shows the oracle used? It can be a link to Grizzly docs or a public announcement

PR #6826, opened 2024-05-06
Add Reserve Treasury and Mainnet Governance's ID

KevinMoll-ls:
New Governance's ID:
ETH+, hyUSD, USDC+, USD3 and rgUSD


KevinMoll-ls:
Hello,
It seems to be a problem with the Governance tally option to switch between tokens, is not working at the moment, it shows the following error:
'Application error: a client-side exception has occurred (see the browser console for more information).'

https://defillama.com/protocol/reserve-protocol#governance

KevinMoll-ls:
This problem still persists

PR #6832, opened 2024-05-06
Update kim protocol to add token

agileurbanite:


agileurbanite:
@dtmkeng can we get help getting this merged in

PR #7052, opened 2024-05-27
add based-chad metadata

basechad:


realdealshaman:
Hey @basechad have you already submitted an pull request with the code that returns your project's TVL?

basechad:
@realdealshaman Hey, my only goal here is to update the token logo on your site.  I followed instructions saying a PR should additionally be submitted here so I did that.  My understanding is that an adapter is not necessary for our project (just memes) so I wasn't going to worry about that.

PR #7122, opened 2024-06-04
add Physica Finance and corresponding Planq tokens

ivanshukhov:


ivanshukhov:
Thanks for already merging Physica itself!
Wondering if this is the correct spot to add the additional tokens to be tracked?

PR #7129, opened 2024-06-05
add new pyth users

mariobern:
gm

requesting your review to add Pyth users

Flat Money
https://docs.flat.money/pyth-network-price-oracle 

Superposition
for context, Superposition is powered by Concordia, which uses Pyth for its pricing engine

https://docs.concordia.systems/platform-spec/pricing-engine 

Also there seems to be an issue with Zerolend. It's using Pyth for zkSync but can't seem to find it on the Pyth page https://defillama.com/oracles/Pyth

Could you take a look please?

Thank you very much


realdealshaman:
Hey @mariobern thanks for the pr. Regarding Superposition, we generally don't add Chainlink as the oracle securing the TVL of projects that use AAVE. Similarly, we'd prefer to only add Pyth as the oracle for projects that specifically use Pyth themselves. Can you modify the pr to exclude Superposition at this time? If Superposition directly integrates Pyth in the future, we'd be happy to update the information accordingly

mariobern:
Thank you for the reply sir

Will remove Superposition for now. For Zerolend, do we need to do anything or could you take care of that?

PR #7273, opened 2024-06-17
Update steer.ts, add bsc and base support

derekbar90:


waynebruce0x:
We're currently getting 'Error: bad indexers' for arbitrum and polygon - do you know anything about this?

derekbar90:
Interesting, we haven't seen any of these issues. It's most likely TheGraph related with the changeover.

derekbar90:
We currently use these same graph endpoints in production

PR #7300, opened 2024-06-20
add macaron

Macaromswap:


waynebruce0x:
can price be derived using on chain calls instead of reading from a static json?

waynebruce0x:
closing for inactivity

PR #7318, opened 2024-06-21
chore: update shapeshift description

firebomb1:
Updates the description of the ShapeShift project. The current description mentions "free" which is still accurate for transactions under $1000, but above this amount ShapeShift takes fees now. We've also added services like loans (provided through ThorChain), so we figure it's worth mentioning it too.


firebomb1:
Thanks for the quick review/merge!

PR #7341, opened 2024-06-25
Add Vega Protocol as a Pyth oracle user

jeremyletang:
Adding Pyth as an oracle source for vega protocol.

For info this is a link to the documentation: https://docs.vega.xyz/mainnet/tutorials/using-data-sources#using-a-pyth-price-feed

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -6355,7 +6355,7 @@ const data3: Protocol[] = [
     module: "vega-protocol/index.js",
     twitter: "vegaprotocol",
     forkedFrom: [],
-    oracles: [],
+    oracles: ["Pyth"], // https://docs.vega.xyz/mainnet/tutorials/using-data-sources#using-a-pyth-price-feed
     listedAt: 1684246282,
     github: ["vegaprotocol"]
   },


Define101:
That's only to submit a proposal to add a new market right? How many markets are using Pyth?

Define101:
image
can see some markets using, but not sure why on gnosis

Define101:
https://explorer.vega.xyz/markets/4e9081e20e9e81f3e747d42cb0c9b8826454df01899e6027a22e771e19cc79fc

jeremyletang:
@Define101, yes the link shows how user can submit markets using the pyth oracles (all market on vega are submitted by the community via governance, and they can choose what oracle to use).

Most markets are using Pyth as a oracle price source for funding (as a matter of fact all PERPS markets does). 

The Pyth prices are pushed on Gnosis Pyth contract correct, the Vega networks pull prices from the gnosis chain (every 1 minutes) to be used as external index price for the funding payments. Vega is compatible with any EVM based chain, but Gnosis is used in most cases here as it's the cheapest chain in term of gas, and that allows the community to push Pyth prices update at a high frequenty for very little gas.

Hope that answers those question.

PR #7358, opened 2024-06-26
Update oracles for Term Structure

imfeng:
Hi,

We would like to submit an update oracles information for the Term Structure. Below are the details of the changes:

Oracle Provider(s): RedStone

Implementation Details: Term Structure now utilizes RedStone Price Feeds for determining the values of loans collateralized by LRTs and LSTs

Documentation/Proof:
[https://docs.ts.finance/protocol-spec/primary-markets/liquidation-mechanism#fully-on-chain-liquidation](https://docs.ts.finance/protocol-spec/primary-markets/liquidation-mechanism#fully-on-chain-liquidation)

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -46871,7 +46871,7 @@ const data3: Protocol[] = [
     module: "term-structure/index.js",
     twitter: "TermStructLabs",
     forkedFrom: [],
-    oracles: ["Chainlink"], // https://docs.ts.finance/protocol-spec/primary-markets/liquidation-mechanism#fully-on-chain-liquidation
+    oracles: ["Chainlink", "RedStone"], // https://docs.ts.finance/protocol-spec/primary-markets/liquidation-mechanism#fully-on-chain-liquidation
     audit_links: ["https://github.com/term-structure/audits"],
     github: ["term-structure"],
     listedAt: 1718906602


realdealshaman:
Hi @imfeng, thanks for the PR! How much of the TVL do loans using LRTs and LSTs as collateral make up? Additionally, if one of these were hacked, would 50% of the Term Structure TVL be at risk?

imfeng:
Hi @realdealshaman 

Thank you for your questions. Currently the loans TVL using LRTs and LSTs are almost 0.
In case the TVL of the loans using LRT/LST as collaterals increase significantly and any LRT or LST protocols are hacked, 50% or any percentage of Term Structure TVL will be at risk depending on the collateral pool at the time. This situation is similar to that of other big protocols such as AAVE would face.

imfeng:
Hi @realdealshaman ,
Since we operate a peer-to-peer lending system, only those lenders who accept LRT/LST as collateral will bear these risks. This means that the risks associated with general collaterals (e.g., USDT, ETH, WBTC, etc.) are segregated from those associated with LRT/LST collaterals.

If an anomaly such as a hack on LRT/LST occurs, resulting in insolvency, the impact can be limited to the lenders who are willing to take on these risks.

I hope these additions are satisfactory. Please let me know if there's any other information you need, and how we might proceed with the approval of these changes.

realdealshaman:
Hey, @imfeng  thanks for the info. Unfortunately, we don't separate TVL per token, so we'll need to wait until LRT/LST collaterals make up more than 50% of the entire TVL for the protocol before we can add it to Redstone. Otherwise, the TVL from other collaterals would be unfairly attributed to the oracle

imfeng:
> Hey, @imfeng thanks for the info. Unfortunately, we don't separate TVL per token, so we'll need to wait until LRT/LST collaterals make up more than 50% of the entire TVL for the protocol before we can add it to Redstone. Otherwise, the TVL from other collaterals would be unfairly attributed to the oracle

I understand, thank you for your response.
I close this PR

imfeng:
> Hey, @imfeng thanks for the info. Unfortunately, we don't separate TVL per token, so we'll need to wait until LRT/LST collaterals make up more than 50% of the entire TVL for the protocol before we can add it to Redstone. Otherwise, the TVL from other collaterals would be unfairly attributed to the oracle

I understand, thank you for your response.
I close this PR

PR #7372, opened 2024-06-27
Important update to Balanced information

usr-icon-foundation:
**Accurately naming Balanced and it's two underlying products (DEX + Stablecoin)**

### Parent protocol

- Name: Balanced
- Description: Balanced is DeFi designed for adoption: it’s fast, affordable, and easy to use. Home to the Balanced Dollar stablecoin (bnUSD) and a decentralised exchange, it uses ICON’s cross-chain technology to connect to other blockchains. You can use Balanced to borrow bnUSD, swap assets, supply liquidity, and transfer cross-chain. The best part is, your crypto stays wrapper-free.
- Website: https://balanced.network/
- Methodology: Fees: Collected from borrowers and traders. TVL: The total liquidity held on the Balanced exchange and used as collateral for bnUSD

---

### Child protocol: DEX

- Name: Balanced Exchange
- Description: Balanced includes a decentralised cross-chain exchange. Use it to swap assets, supply liquidity, and transfer crypto between supported blockchains. Its unified approach to cross-chain assets removes the need for wrapped tokens and fragmented liquidity, so you can enjoy a simpler trading experience.
- Website: https://app.balanced.network/trade
- Methodology: TVL: The total liquidity held on the Balanced exchange and in the Stability Fund. Fees: Collected from traders

---

### Child protocol: Stablecoin

- Name: Balanced Dollar
- Description: The Balanced Dollar (bnUSD) is a decentralised, cross-chain stablecoin that tracks the price of 1 USD. It uses cryptocurrency as collateral to guarantee its value, with support for assets from a variety of blockchains. bnUSD is not experimental or algorithmic: it’s over-collateralised, so the total supply cannot exceed the value that backs it. You can borrow bnUSD, or buy some from the Balanced exchange. If you have bnUSD sitting idle in your wallet, deposit it into the Balanced Savings Rate to earn a return.
- Website: https://balanced.network/stablecoin
- Methodology: TVL: The total liquidity deposited as collateral and held in the Stability Fund

PR #7391, opened 2024-06-27
fix commit #7372

Define101:
fix commit #7372


usr-icon-foundation:
You have explained before, that each child protocol must carry the ticker of the parent protocol.

**defi/src/protocols/data3.ts** | Balanced Dollar

- In this case it makes no sense for $bnUSD stablecoin to carry the $BALN ticker. What edits could we work on, for the bnUSD stablecoin to carry its own ticker bnUSD?

Thanks! David

PR #7419, opened 2024-07-03
Adding Celo stablecoin addresses

marekolszewski:
The stablecoin section is showing incorrect data for the Celo chain. 

Adding 5 stablecoin addresses that collectively have a TVL of around $240M.

USDC contract address taken from:
https://tether.to/en/supported-protocols/

Tether contract address taken from:
https://tether.to/en/supported-protocols/

Celo stablecoin contract addresses taken from:
https://docs.celo.org/token-addresses

### DIFF
--- a/coins/src/adapters/tokenMapping_added.json
+++ b/coins/src/adapters/tokenMapping_added.json
@@ -3778,6 +3778,36 @@
     }
   },
   "celo": {
+    "0xcebA9300f2b948710d2653dD7B07f33A8B32118C": {
+      "name": "USD Coin",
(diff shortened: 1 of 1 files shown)


marekolszewski:
As an aside, is there something else that we need to change to make this tokens show up here:

https://defillama.com/stablecoins/Celo

realdealshaman:
Hey @marekolszewski thanks for the PR! I've added the coins to https://defillama.com/stablecoins/Celo, it will be visible with the next hourly update. We already had celo euro and celo dollar but I have now also added celo real too

PR #7394, opened 2024-06-28
Update images for Balanced Exchange + Balanced (parent protocol)

usr-icon-foundation:

PR #7476, opened 2024-07-09
commit #7394

Define101:


usr-icon-foundation:
Thanks!

usr-icon-foundation:
Hi @Define101 I have noticed that the Stablecoin Info in incorrect. 

It seems like the wrong stablecoin is being referenced here. The Balanced Dollar stablecoin goes by **bnUSD**.  

David

image

PR #7518, opened 2024-07-13
update YEL token address and info after migration

yeloper:
update YEL token address and info after migration


realdealshaman:
Hey, @yeloper thanks for the pr but we can't add new chains to the listing unless they are also in the tvl adapter. Please update this adapter to add new chains: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/yel/index.js

yeloper:
Already update new chains, check pls

realdealshaman:
@yeloper thanks but I see that only Blast has been added to the TVL adapter. There are also a few fields that we can't change like the listedAt for example. Should we also have the Name change now completely so that it is no longer Yel Finance?

yeloper:
@realdealshaman Yes, I see that you have adjusted the data files. This will be correct for now.

PR #7579, opened 2024-07-22
Update oracles for Yei Finance and Bitlen Finance.ts

Hundert1000:
Hey Llamas,
Kindly asking to update oracles for Yei Finance on Sei and Bitlen Finance

Oracle Provider(s): RedStone

Implementation Details: Yei Finance uses RedStone USDT, USDC feeds as primary which is >50% of their TVL. This info is clearly stated in their docs ("RedStone USDT and USDC push feeds are used as primary oracle for these assets.") and confirmed by their team.

Bitten Finance uses RedStone as the primary solution for asset pricing.

Documentation/Proof:
[Yei Finance] https://docs.yei.finance/welcome-to-yei/security-and-risk/oracles-and-data-feeds
[Bitlen Finance] https://docs.bitlen.io/borrowing/oracle

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -41613,7 +41613,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Lending",
     chains: ["BSquared"],
-    oracles: ["Supra"], //https://supra.com/docs/data-feeds/pull-model/networks/
+    oracles: ["RedStone"], //https://docs.bitlen.io/borrowing/oracle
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @Hundert1000 thanks for the pr, looking it Yei Finance we can see that the way their oracle works is by checking pyth first and if the price is stale it will check RedStone, in this case we'd keep Pyth as the primary oralce. In the case of Bitlen, we can update it with your changes. Please update the PR to apply changes only to the Bitlen listing and we can merge. Thanks again

PR #7658, opened 2024-08-05
Update oracles and chains for Primex Finance

Mar-Alex:


realdealshaman:
Hey @Mar-Alex thanks for the PR! I see that only Pyth is mentioned in their docs. Please update it to only Pyth and add the reference link https://docs.primex.finance/guides/what-is-primex/economic-security/oracles/pyth-network-oracle and we'll get it merged

Mar-Alex:
Hi @realdealshaman 
We've added information about how we use Umbrella and Chainlink in the protocol to the documentation
https://docs.primex.finance/guides/what-is-primex/economic-security/oracles/chainlink
https://docs.primex.finance/guides/what-is-primex/economic-security/oracles/umbrella

realdealshaman:
Thanks for the additional info, @Mar-Alex ! Can you tell us which is your primary oracle? To add the correct oracle that secures the protocol's TVL, we need to know: If the oracle gets hacked, would 50% of the protocol's TVL be at risk? We need to ask this for Chainlink and Pyth, but it seems this is not the case for Umbrella, as it is just a gas TWAP for keepers

Mar-Alex:
@realdealshaman Currently, the security of the protocol mostly depends on Chainlink, but we are switching to Pyth to increase the number of supported pairs. If CL or Pyth is hacked, the protocol TVL would be at risk.

Regarding Umbrella, you are right, they don't protect TVL directly, it's more about correct keeper reward calculation.

Mar-Alex:
@realdealshaman Is it possible to mention several oracles, as we depend on all of them, or at least both Chainlink and Pyth? Pyth's importance for the protocol will expand rather fast during the next weeks.

realdealshaman:
@Mar-Alex  We need to only mention the oracles securing the TVL. We can add more than one but only once Pyth is securing a significant part of the TVL, otherwise the total amount of the TVL will also be attributed to them incorrectly

realdealshaman:
I've  added Chainlink as per the comment above, please let us know when Pyth is more widely used and can be considered the primary oracle and we can update the listing.

PR #7666, opened 2024-08-05
Updating migrated token address, URL and other info

CryptoBKT:
updated to migrated token address, updated from old site URL, updated description for new products, updated to appropriate category, updated number of audits, updated chains.

Ref: https://docs.acryptos.com/dao/acs


realdealshaman:
Hey, @CryptoBKT thanks for the PR! Have you also applied changes to the tvl adapter so that it tracks the tvl of the Liquidity manager product instead of the yield aggregator?

CryptoBKT:
> Hey, @CryptoBKT thanks for the PR! Have you also applied changes to the tvl adapter so that it tracks the tvl of the Liquidity manager product instead of the yield aggregator?

Thanks for the reply! Yes, we have been updating the TVL adapter for the liquidity manager all this while, but we didn't update these information-based sections up till now.

Migration from old token address wasn't updated as well.

PR #7691, opened 2024-08-09
Update aspida info

Tiger-Aspida:


Define101:
@Tiger-Aspida 
What is the difference between aETH and saETH? Are they both redeemable 1:1 with ETH?

Tiger-Aspida:
Here’s how you can respond:

aETH is a token that maintains a 1:1 ratio with ETH, minted from either ETH or stETH, and can be held in a user’s address or further staked to become saETH. The exchange rate between aETH and saETH is implemented according to the ERC4626 standard and is subject to change over time, as it is based on linear yield distribution. The yield sources include underlying native staking rewards, stETH rewards, and EigenLayer restaking rewards. Therefore, while aETH is redeemable 1:1 with ETH, the exchange rate between aETH and saETH is not fixed and will vary over time.

Tiger-Aspida:
@Define101 Can you merge this PR?

Lobonegro8 (approved):

PR #7790, opened 2024-08-20
Fix Random Reward

roinevirta:
Fixes description & category.


realdealshaman:
@roinevirta thanks for the PR! We’d like to keep the current category as "Prediction Market" since we try to avoid using the "Services" category. Even though the result is random, users are buying a future outcome, which aligns with the "Prediction Market" definition: protocols that allow users to wager, bet, or buy into future results

PR #7797, opened 2024-08-20
Added Token to tokenMapping_added.json for Sei

pho-phoenix:
- added FRAX, frxETH, sfrxETH, sFRAX which are bridged by Stargate
- added JLY the native token from Jellyverse which is the official friendly fork from Balancer on Sei


waynebruce0x:
in https://github.com/DefiLlama/defillama-server/pull/7805 I have moved these mappings to tokenMapping.json, which is the file we use for new mappings.

PR #7914, opened 2024-09-02
Update Dinari price decimals

jaketimothy:
This aligns with a protocol upgrade to fix price feed decimals


waynebruce0x:
is this change still required in light of the recent work from @g1nt0ki ? https://github.com/DefiLlama/defillama-server/blob/master/coins/src/adapters/rwa/dinari.ts

jaketimothy:
> is this change still required in light of the recent work from @g1nt0ki ? https://github.com/DefiLlama/defillama-server/blob/master/coins/src/adapters/rwa/dinari.ts

Yes, this is still needed

PR #8004, opened 2024-09-12
Removed Oath Foundation parent from Harbor.market

Deantoshi:


Deantoshi:
The Harbor team is a licensed fork of the Oath Foundation, but requested that Oath not be listed as a parent.

PR #8016, opened 2024-09-13
fix: lombard coin gecko and category

jameswstevens:
Added Coin Gecko.

Moved to liquid staking. Lombard is a Bitcoin Liquid staking platform. While users can move BTC liquidity cross-chain, the main attribute of Lombard is our liquid staking token (LBTC).

We are the only pure BTC LST, which makes it a bit harder to classify us since we are both clearly adopting BTC liquidity, but more importantly staking BTC.

The defi community sees us as liquid staking if this helps align :) :

https://x.com/Pell_Network/status/1834583754477891623
https://x.com/StakingRewards/status/1833834254599426181
https://x.com/ZircuitL2/status/1833536163468218765
https://x.com/MitosisOrg/status/1833520897573196164
https://x.com/symbioticfi/status/1831754460986785856

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -54354,9 +54354,9 @@ const data3: Protocol[] = [
     logo: `${baseIconsUrl}/lombard.jpg`,
     audits: "2",
     audit_note: null,
-    gecko_id: null,
+    gecko_id: "lombard-staked-btc",
     cmcId: null,
-    category: "Bridge",
+    category: "Liquid Staking",
     chains: ["Bitcoin"],
     module: "lombard/index.js",
     oracles: ["RedStone"], //https://github.com/DefiLlama/DefiLlama-Adapters/pull/11458


Define101:
There is no staking of BTC. 
Category should be bridge. Since 1 LBTC = 1 btc "according to your website"
And regarding the coingecko id, we only add it when is the governance token

jameswstevens:
Thank you for clarification on gov. token. 

Regarding BTC staking -

As of a few weeks ago, Babylon (founded by Professor David Tse of Stanford) unlocked the first every BTC staking mechanism. Using a timelock tx and Shnorr sIgnatures, BTC holders are now able to have slashable validation on Cosmos PoS chains backed by this locked BTC on the Bitcoin chain.

At Lombard, users provide BTC, which we stake the max amount (limited currently by Babylon) into the Babylon staking timelock tx. In early October, this will increase to 1,000BTC and by next year we will have all BTC staked. Since this BTC is staked into Babylon and we, in turn, provide LBTC for each BTC locked, this should technically be considered a Liquid Staking protocol.

Let me know if you would like to chat more about it! It is very new, so I understand why it might be hard to categorize at the moment. :)

jameswstevens:
@Define101. I know you have to be extremely busy, but bumping here :)

Define101:
> At Lombard, users provide BTC, which we stake the max amount (limited currently by Babylon)

Does 100% of the BTC provided by the users is used/deposited into Babylon?

As of now, all btcs will be categorized as the same. The main concern is to provide information regarding the collateral behind those "btc" since most of them are actually not backed by BTC..


Also due to lack of transparency by the involved parties, we need to address this in a slower way, by manually confirming everything (collateral of the respective btcs). Once that process is more clear/transparent, and everything is correct, we can then create a new category for some of this BTCS. Since there are a few types. As of now, the only BTCs that have the category Liquid Restaking are the ones that the collateral asset is WBTC/TBTC staked via eigen or symbiotic (this is real restaking) on ethereum

Define101:
And with all respect, just because someone create a chain and use worthless tokens as rewards, to attrack BTC, does not change the fact that the real bitcoin network, the mechanism is POW. Just because your BTCs can be slashed does not change the fact that bitcoin is not a POS chain.

Also with Babylon particular case, there are different rules for different teams, some teams use 50000000000 wallets others use 2.  Even though the limit is 0.005 BTC. A few teams use 500000 wallets with 0.005, while others use those same 500000 wallets but than transfer those funds to particular wallets like 4 or 5 wallets (making it easy to track it). So yeah, its all shady and untransparent.

Define101:
0.05 BTC sorry typo

PR #8040, opened 2024-09-17
Add BFBTC

lukeInAsd:
Add BFBTC price


waynebruce0x:
Can you explain how BFBTC is pegged to BTC?

lukeInAsd:
Hi , We have checked the amount of  collateral BTC and  bBTC  with AIlayer team,
and we have  burned the corresponding amount of BFBTC, keep the same with bBTC supply.

PR #8050, opened 2024-09-18
update Jupiter perps oracle

mariobern:
gm llama sirs,

a small update on the oracle section of Jupiter perps 

Jupiter perps currently leverages their own oracle (built with Edge) to update prices and open/close positions

However, Jupiter perps still relies on Pyth oracle:

- As a reference price check (sanity check) against their oracle, ensuring that the deviation is not too big.
- As a fallback price if their oracle's prices are stale.

In other words, if e.g. Jupiter's oracle price deviates too much from the Pyth oracle, there will be no trades executed on their perps

This signifies that the Pyth oracle is still a critical part of the Jupiter perps infrastructure as a supporting oracle

https://station.jup.ag/guides/perpetual-exchange/how-it-works#oracle

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -31315,7 +31315,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Derivatives",
     chains: ["Solana"],
-    oracles: ["Edge"], // https://station.jup.ag/guides/perpetual-exchange/how-it-works#oracle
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey, @mariobern would the funds be at risk if Pyth reported an incorrect price?

realdealshaman:
Closing this for now since docs indicate Pyth is a secondary/fallback oracle

PR #8082, opened 2024-09-23
throw error for bad block timestamps

waynebruce0x:


coins/src/getBlock.ts
```diff
@@ -47,23 +47,20 @@ function zkSyncBlockProvider() {
...
 export const blockPK = (chain: string) => `block#${chain}`
 
-async function getBlock(provider: any, height: number | "latest", chain: string): Promise<TimestampBlock> {
+async function getBlock(provider: any, height: number | "latest", chain: string): Promise<TimestampBlock | IResponse> {
   const block = await provider.getBlock(height)
   if (block === null) {
-    throw new Error(`Can't get block of chain ${chain} at height "${height}"`)
+    return errorResponse({ message: `Can't get block of chain ${chain} at height "${height}"`})
+  }
+  const previous = await getClosestBlock(blockPK(chain), block.timestamp - 100, 'low')
```
0xngmi, line 55: why timestamp-100? id do timestamp-1

waynebruce0x: fixed in next commit

0xngmi (approved):

PR #8104, opened 2024-09-25
change category to Options

WayneAl:


realdealshaman:
Hey @WayneAl based on the description for DOV, the best category would be "Options Vault". This is because the protocol allows users to deposit collateral into an options strategy

PR #8113, opened 2024-09-26
Change Extra Finance's category

ExtraFi:
Change Extra Finance's category to Lending.


realdealshaman:
Hey @ExtraFi the category can't be lending unless you are subtracting a borrowed amount from a supplied amount and also exporting the borrowed total as borrowed tvl

PR #8138, opened 2024-09-30
Update SolvBTC oracles

buchaoqun:
Dear DeFiLlama team,

We submit a PR adding RedStone as an Oracle protecting Solv TVS after the integration of solvBTC price feed internally for stability mechanism.
RedStone determines the market price of slovBTC for our minting and redeeming process. This is a core mechanism for the stability of our flow. If the price were to be corrupted, a malicious actor could drain the value from the system.


Define101:
We added here https://github.com/DefiLlama/defillama-server/pull/8151

PR #8148, opened 2024-09-30
Update description Summer.fi

biancabuzea200:
Adding Chronicle as powering Summer.fi
Summer.fi uses Chronicle's Medianizer, as indicated in their [docs](https://docs.summer.fi/protocols/sky-ecosystem/oracle-security-module)

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -46327,7 +46327,7 @@ const data3: Protocol[] = [
     module: "summer-fi/index.js",
     twitter: "summerfinance_",
     forkedFrom: [],
-    oracles: [],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @biancabuzea200, would funds in Summer.fi be at risk if the medianizer were compromised or reported incorrect data?

biancabuzea200:
Hi, @realdealshaman. Yes, it would be at risk.

realdealshaman:
Hi @biancabuzea200, since Chronicle is securing MakerDAO instead, the positions within automation, and its TVL is already counted under Chronicle, adding Summer.fi separately isn't necessary

PR #8202, opened 2024-10-08
feat: update APX Finance oracles

bryansniyve:
Proof：
https://medium.com/apx-finance/apx-v2-integrates-with-pyth-network-for-enhanced-real-time-price-accuracy-and-security-31d0ba5fb2b0

image
This says that Pyth is a primary oracle, chainlink is backup oracle.You can also check with pyth, this information is visible in the contract

You can also check this: https://www.apollox.finance/en/futures/v2/BTCUSD (Pyth logo on APX website)

btw, you can also check with Pyth team

Thanks very much!

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -9135,9 +9135,7 @@ const data2: Protocol[] = [
     cmcId: "16334",
     category: "Derivatives",
     chains: ["Binance"],
-    oraclesByChain: {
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey, thanks for the clarification, in all projects we are looking only to have the primary oracle added to the listing, leave Pyth as the oracle that secures the tvl of APX Finance

PR #8231, opened 2024-10-10
Add Redstone, Pyth and Chainlink oracles for Euler v2

MickdeGraaf:
Added oracles used in Euler v2 markets

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -53168,7 +53168,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Lending",
     chains: ["Ethereum"],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @MickdeGraaf thanks for the PR! Where can we find info on how these oracles are securing the TVL? Which is the primary oracle?

MickdeGraaf:
All vault(lending market) creators can pick their own pricing.

We use them to price collateral and debt.

https://app.euler.finance/vault/0x82c710B9B225B43C0c90f097DB31f5f10a7F86FA on the collateral section you can see which oracles this vault uses for example.

oracles.euler.finance shows all of the adapters euler supports from all different providers.

PR #8248, opened 2024-10-11
update: Cyberperp data update (logo and fields)

eriscrypt:
Update existing Cyperperp data


realdealshaman:
Hey @eriscrypt I've updated the cyberperp info, I see that the logo was updated previously. New info will show in about one hour

PR #8318, opened 2024-10-18
Added Echo aBTC on Aptos

echo-pro-dev:
https://www.coingecko.com/en/coins/abtc


Define101:
We are already pricing that asset here
https://coins.llama.fi/prices/current/aptos:0x4e1854f6d332c9525e258fb6e66f84b6af8aba687bbcb832a24768c4e175feec::abtc::ABTC?searchWidth=4h

PR #8415, opened 2024-10-29
worldes rwa token commit

will2024:


coins/src/adapters/moneyMarkets/worldes/worldes.ts
```diff
@@ -0,0 +1,48 @@
+import { Write, } from "../../utils/dbInterfaces";
+import { getApi } from "../../utils/sdk";
+import { addToDBWritesList, getTokenAndRedirectDataMap } from "../../utils/database";
+
+export const config = {
+    arbitrum: {
+        lps: [
+            '0x80c89197ba29c0ad921ad0f7c2e1403579d36c65',
```
waynebruce0x, line 8: where can we see how much liquidity is in this lp? why is this contract used instead of the one shown on the UI (0x378B6a8bAe0767B714dF136FABE388DfcC623352)?

waynebruce0x:
and how can I verify the results without an invite code for your ui

coins/src/adapters/moneyMarkets/worldes/worldes.ts
will2024: We can obtain LP from the website and verify: https://app.worldes.io/liquid-pool

PR #8417, opened 2024-10-29
Reclassify Puffer Oracle

lelouch-crypto-0:
Hello Defillama team, I took the time to do some due diligence for this project.

According to the [docs](https://docs.puffer.fi/protocol/validator-tickets#pricing-validator-tickets) the pricing is done via Redstone oracles for the VTs but upon further investigation the current pricing mechanisms of VTs happen via a primary check of gaurdians who choose the redstone price that is posted at the 12 hour interval.

What this means is that a redstone misreport will not be accepted by the guardian because they would choose not to post it. According to the definition of [TVS](https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254)  and the question "If the oracle malfunctioned, would those assets be lost?" the answer here is no

I checked with the puffer team and they agree with my conclusion

ishare-1730197922

ishare-1730198056

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -31577,7 +31577,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Liquid Restaking",
     chains: ["Ethereum"],
-    oracles: ["RedStone"], // https://github.com/DefiLlama/defillama-server/pull/7986
+    oracles: [""],
     forkedFrom: [],
     module: "puffer/index.js",
     twitter: "puffer_finance",
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -31577,7 +31577,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Liquid Restaking",
     chains: ["Ethereum"],
-    oracles: [""],
+    oracles: [],
     forkedFrom: [],
     module: "puffer/index.js",
     twitter: "puffer_finance",


JasonVranek:
Chiming in here as a core contributor at Puffer to remedy this. I believe our community moderator was speaking from the previous iteration of the protocol before the Redstone integration.

Today, Redstone is calling setDailyRewardsAndPostMintPrice (https://github.com/PufferFinance/puffer-contracts/blob/3be1b03e367a6ee4e3a252f4bc9ae8865ff7d42f/mainnet-contracts/src/ValidatorTicketPricer.sol#L101) which will update the VT price without intermediates. In the code there are some guardrails here (https://github.com/PufferFinance/puffer-contracts/blob/3be1b03e367a6ee4e3a252f4bc9ae8865ff7d42f/mainnet-contracts/src/ValidatorTicketPricer.sol#L190) and here (https://github.com/PufferFinance/puffer-contracts/blob/3be1b03e367a6ee4e3a252f4bc9ae8865ff7d42f/mainnet-contracts/src/ValidatorTicketPricer.sol#L207) to prevent extreme outlier values from being accepted and adversely affecting the protocol; however, that is enforced atomically at the smart contract level and not by the guardians.

Define101:
@lelouch-crypto-0 who made those contracts?

lelouch-crypto-0:
@JasonVranek thanks for correcting and good to see that the price is being used atomically. However the point still stands whether the price being used atomically or via gaurdians that the oracles don't actually "secure" the protocol. The definition of [TVS](https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254) and the question "If the oracle malfunctioned, would those assets be lost?" is still a no because any large misreport by redstone would easily just be ignored since the tolerance is current set at 5%  ( for consensus payouts), can be verified [here](https://etherscan.io/address/0x9830aD1bD5Cf73640e253EdF97DeE3791C4a53C3#readContract). 

given above info Attributing 1.5 Billion worth of assets seems disingenuous and I can't help but feel that people don't understand how the convention works.

Also according to the validator pricing history [here](https://etherscan.io/txs?a=0x9830aD1bD5Cf73640e253EdF97DeE3791C4a53C3&p=2) only 1 address has ever called the `Set Daily Rewards And Post Mint Price` function which is this address: [0x65d2dd7A66a2733a36559fE900A236280A05FBD6](https://etherscan.io/address/0x65d2dd7a66a2733a36559fe900a236280a05fbd6) . Give the transaction history of this address it clearly looks like an EOA that belongs to the puffer team so I fail to see where redstone oracle transactions originate from. If you can share that it would be helpful but regardless it still doesn't satisfy TVS conventions so shouldn't be attributed.

lelouch-crypto-0:
@Define101 don't understand the question, the contracts are created by the puffer team but the attribution of TVS shouldn't be left to the project submitting the information because the due diligence isn't there. I've given my rationale in the above comment as to why this is false attribution of TVS.

Define101:
Sorry wrong ping @lelouch-crypto-0 . Wanted to ask to @JasonVranek

jakub-wojciechowski:
Hey, @Define101 as we're the ones implementing the mechanism, happy to share more details and answer your concerns:
1. The prices is submitted directly through [redstone oracle contract](https://etherscan.io/address/0x604dba5dba0dd58997dcf9b5ae5d066c8e14a3c0). It is not listed on etherscan in transactions tab because it is “internal transaction” submitted transactions can be observed in [“internal transactions”](https://etherscan.io/address/0x9830ad1bd5cf73640e253edf97dee3791c4a53c3/advanced#internaltx) or in [events tab](https://etherscan.io/address/0x9830ad1bd5cf73640e253edf97dee3791c4a53c3/advanced#events)
The flow looks like this:
1) Redstone relayer send a transaction to Redstone validation contract https://etherscan.io/address/0x9830ad1bd5cf73640e253edf97dee3791c4a53c3/advanced#internaltx
2) Redstone validation contract then directly calls Puffer ValidatorTicketPricer 0x9830aD1bD5Cf73640e253EdF97DeE3791C4a53C3
2. There are not any intermediaries or guardins involved.The argument of having 5% tolerance as sufficient prevention against missprice is very short-sighted. For example, redstone-oracles could submit a slightly deviated price by 4.99% in every block. And after 10 blocks achieve 4.99 *10 = 49.9% deviation, which could lead to Validator Ticker price market collapse.
BTW, @lelouch-crypto-0, if you've got any concerns regarding the Oracle mechanics during your research, I'm always available on our Discord (even via DM's) or you can ping any of the other team members.

lelouch-crypto-0:
thanks @jakub-wojciechowski regarding the 2 points namely,

1. I can indeed see that `Set Daily Rewards And Post Mint Price` being called by redstone oracle in the internal transactions.
2. Yes I agree, after examining the code I can confirm that the above is possible if the oracle is malicious since apart from the gaurdrail of 5% there isn't any other gaurdrail so that theoretical possibility is there.

However, I took the time to examine the entire protocol and the code to answer the question if redstone is actually securing the ETH-pufETH assets by pricing Validator tickets and the answer is I arrived at is **it isn't**. Let me explain how the protocol works in simple terms and the role of validator tickets for everyone here to understand.

this diagram explain how the protocol works: 

image

1. Stakers deposit ETH into the [PufferVault](https://etherscan.io/address/0xD9A442856C234a39a81a089C06451EBAa4306a72) 
2. Puffer Vault mints pufETH to ETH Depositor which accumulates yield from restaking 
3. Node Operator deposit 28 Validator tickets + 1~2 ETH to participate in restaking . Atm it costs roughly [100 dollars for 15 tickets](https://etherscan.io/tx/0x4dba5cbf201bc4b9a14ed18b858e969e88c3107ef72233b3f26a0feef2e2171f) 
4. Vault funds PufferModule with 32 ETH
5. puffer protocol delegates restaking to Node Operator
6. restaking awards are distributed to pufETH holders and PoS awards goes to the node operators

lets consider the role of validator tickets, their main purpose is to allow node operators to be able to register as a validator within the puffer protocol, 1 ticket represents one validator-day of expected PoS rewards. The payments to mint VTs directly pay pufETH holders. Apart from this VTs can also be traded on secondary markets since they are ERC20s. 

Now, lets take 2 scenarios of a malicious oracle attack:

1. The oracle increases the price of VTs by 5% every 40 seconds ( the limit set within redstone contracts) to something very very large
2. The oracle decreases the price of VTs by 5% every 40 seconds to something close to 0 (can never reach 0)

In Scenario 1 the only effect this would have is that no new node operators would be able to join as validators. This is trivial because the gaurdians would remove the malicious oracle and start posting the correct prices having no effect on puffer whatsoever.

In Scenario 2 , someone would be able to mint a lot of VTs before the protocol/guardian notices that the price was set incorrectly and remove the malicious oracle and set the correct price.

What effect would minting a lot of VTs have on puffer? The biggest effect it would have is that they could attempt to join as restaking validators within the puffer protocol but this would be easily identifiable and the gaurdians can choose to not [provision them](https://docs.puffer.fi/protocol/guardians#what-are-their-duties) or in the worst case if they do manage to get provisioned, the guardians can [eject them](https://docs.puffer.fi/protocol/guardians#what-are-their-duties) . In the very minuscule chance that the guardians provision rogue node operators and fail to eject them the [Burst threshold](https://docs.puffer.fi/protocol/burst-threshold) would be hit which will not allow more validators to join.

The [exchange rate](https://docs.puffer.fi/protocol/nlrt#calculating-the-conversion-rate) of pufETH-ETH depends on the restaking operators in the puffer modules and their performance. The pricing of validator tickets will not have an effect on this unless the gaudians are also comprimised and even if they are comprimised it won't have an immediate effect which will result in >50% TVL being drained. 

The [TVS specification](https://github.com/DefiLlama/DefiLlama-Adapters/discussions/6254) is pretty clear in what its asking, "If the oracle malfunctioned, would those assets be lost?" . Even while assuming a rogue oracle the assets aren't lost ( which shouldn't be needed because a misreport is different from a rogue oracle ) . So if the conditions for the assets to be lost is a rogue oracle and a rogue guardian then this should not qualify under TVS.

I'll let the Defillama team make their best judgement, I've made the best case of why this is false attribution. Initially I considered VT pricing to be important but after doing a deeper dive this is not the case so I was wrong in my first comment. Any oracle posting the VT price will not result in >50% assets being lost due to the design of the protocol doesn't matter if its redstone or guardians.

lelouch-crypto-0:
Sure @jakub-wojciechowski will be happy to ping

JasonVranek:
Apologize for the delay, I've been quite sick. Will still leave it to the Defillama's teams judgement as to whether this satisfies TVS reqs but I want to clarify a few points from lelouch's description. 

> In Scenario 1 the only effect this would have is that no new node operators would be able to join as validators. This is trivial because the gaurdians would remove the malicious oracle and start posting the correct prices having no effect on puffer whatsoever.

I agree with this analysis. But for scenario 2 it is not so easily recoverable:

> What effect would minting a lot of VTs have on puffer? The biggest effect it would have is that they could attempt to join as restaking validators within the puffer protocol but this would be easily identifiable and the gaurdians can choose to not [provision them](https://docs.puffer.fi/protocol/guardians#what-are-their-duties) or in the worst case if they do manage to get provisioned, the guardians can [eject them](https://docs.puffer.fi/protocol/guardians#what-are-their-duties) .

VTs are fungible ERC20s, so it is not as simple as telling the guardians not to provision certain validators or to eject the ones that managed to register with these 'free VTs'. Once the total supply has exploded likely it would require a hard fork of the protocol to instantiate a new VT contract which of course should be avoided at all costs.

> The [exchange rate](https://docs.puffer.fi/protocol/nlrt#calculating-the-conversion-rate) of pufETH-ETH depends on the restaking operators in the puffer modules and their performance.

This statement is actually the exact opposite of the intention of VTs. We implemented them because we wanted to _decouple_ the performance of permissionless validators from the rewards that the LRT holders receive. Ignoring restaking rewards which are out of scope of this convo, the only way the pufETH-ETH conversion rate increases is if validators purchase VTs. If our VT price goes to zero, then validators would be able to borrow staker capital freely, e.g., LRT yield goes to 0 which would obviously result in our TVL going to 0.

0xtechdean:
I believe there's a crucial point missing in this discussion: off-chain data updates inherently require reliance on a trusted source. Just as decentralized applications can generate off-chain price feeds but often choose to rely on trusted third-party oracles for these critical updates, we should adopt a similar approach here. Arguing against the use of trusted oracles in favor of DApps handling off-chain data internally may inadvertently incentivize protocols to become more centralized, which is contrary to the foundational principles of DeFi and should be viewed critically. Critical off-chain updates should be conducted by trusted oracles to protect depositors from potential manipulation—even by internal actors like guardians.

Furthermore, it seems that your argument relies on the guardians noticing significant differences between the actual price of the VT's and the price provided by the oracle. However, the oracle can also manipulate the price subtly over time by making small, incremental adjustments that are less detectable. Even these minor changes, when accumulated over time, can lead to a collapse of the VT market.

Therefore, while the guardians might be effective in preventing attacks that involve obvious discrepancies, they may struggle to detect and respond to more sophisticated attacks where the oracle manipulates prices gradually. In such scenarios, it could become very challenging for the guardians to act as they should and in a timely manner. This potential delay or oversight increases the risk of significant asset loss within the protocol.

In essence, relying solely on guardians to mitigate oracle manipulation may not be sufficient to safeguard the protocol against all forms of attack. By leveraging trusted oracles for critical off-chain updates, we enhance security and maintain the decentralized nature of the protocol, ensuring that no single party has undue influence over essential operations. This approach not only protects depositors from potential manipulation but also aligns with DeFi's commitment to decentralization and trust minimization.

Moreover, advocating to remove this from DefiLlama disincentivizes oracles from doing this important work and could lead to situations where either external entities or teams take advantage of users. Oracles play a crucial role in maintaining the integrity and security of decentralized protocols. By undermining their importance or excluding them from metrics like TVS, we risk diminishing their contributions and potentially compromising the system's overall security. Recognizing and supporting the role of trusted oracles is essential to prevent centralization and protect users from potential exploitation.

lelouch-crypto-0:
Hey @0xtechdean what you've said is partially correct but there are a few key points:

> Furthermore, it seems that your argument relies on the guardians noticing significant differences between the actual price of the VT's and the price provided by the oracle. However, the oracle can also manipulate the price subtly over time by making small, incremental adjustments that are less detectable. Even these minor changes, when accumulated over time, can lead to a collapse of the VT market.

Yes, that can happen but that doesn't mean the protocol TVL is at risk, like @JasonVranek at worst the protocol has to be forked and TVL will need to be migrated but this is entirely different from how an oracle misreport can effect other protocols like AAVE, Perpetual etc, A single misreport there can cause millions in funds to be lost on those protocols. Even in the worst case of puffer ( which is assuming a rogue malicious oracle) the underlying ETH is barely at risk. The arguments of small incremental updates would hold some merit if the VTs were redeemable for ETH but given its just use to provision node operators ( which first have to qualify via guardians) I think the argument that the oracle is "securing" the protocol is a stretch at best and disingenuous to the methodology already established.

jakub-wojciechowski:
It’s a pretty long thread and every time we try to patiently point out technical inaccuracies in all of the random accusations towards RedStone service. It makes me wonder what the true agenda behind an anonymous “researcher” who accidentally targets only RedStone integrations and produces evidence by interacting with non-tech-savvy discord moderators is. 

As there are no new technical concerns left unanswered, let me comment on the latest arguments. Honestly, I feel shocked reading an argument that “at worst the protocol has to be forked”. I still remember the infamous “DAO hack” and the Fork which was used as a remedy that almost wiped out the Ethereum project. After 8 years, it is being viewed as a very controversial decision and I don’t suppose this strategy will be ever used again. Oracles are here for the reason that we won’t ever need to rewrite history by creating a fork. 

In the previous post, a person behind the technical design of Puffer (Jason, CTO) clearly stated that a failure of RedStone Oracle will cause the TVL to drop to 0. In light of that, it’s hard to argue that RedStone doesn’t play an important role there and shouldn’t be attributed to the Puffer protocol. I’d be grateful if @Define101 could kindly reassess the case in the light of all the technical details provided.

lelouch-crypto-0:
@jakub-wojciechowski  this is not directly targeted at redstone whatsoever, I work with a investment firm that I cannot disclose the name for but often times the Defillama dashboard is touted as a source of truth by projects approaching us. Upon investigation I wanted to verify for myself if the dashboard is accurate. 

I will be opening PRs for other oracles and protocols that found to be not accurate just so you are aware.

> In the previous post, a person behind the technical design of Puffer (Jason, CTO) clearly stated that a failure of RedStone Oracle will cause the TVL to drop to 0. In light of that, it’s hard to argue that RedStone doesn’t play an important role there and shouldn’t be attributed to the Puffer protocol. I’d be grateful if @Define101 could kindly reassess the case in the light of all the technical details provided. 

What jason has cited and I have repeated is that this rogue oracle attack result in no loss of ETH within the puffer protocol.  People withdrawing their ETH and the TVL going to 0 because of people withdrawing their ETH is not equivalent of the "funds being lost" which is what the defillama TVS methodology is defined by. The technical inaccuracies you speak of still do not address the simple question of "will the ETH inside puffer protocol be lost if the oracle misreports", the answer is no and can be verified through the architecture of the protocol. 

> As there are no new technical concerns left unanswered, let me comment on the latest arguments. Honestly, I feel shocked reading an argument that “at worst the protocol has to be forked”. I still remember the infamous “DAO hack” and the Fork which was used as a remedy that almost wiped out the Ethereum project. After 8 years, it is being viewed as a very controversial decision and I don’t suppose this strategy will be ever used again. Oracles are here for the reason that we won’t ever need to rewrite history by creating a fork. 

Yes because in the DAO hack 150M dollars worth of Ether was "lost" , emphasis on the word lost. Projects fork or migrate to new versions all the time and in the event that puffer VT price market collapses it shouldn't effect the users of the protocols because funds cannot immediately be lost especially if the guardians never provision them.

PR #8465, opened 2024-11-01
update oracle and audit meso finance

gin-meso:


realdealshaman:
Hey @gin-meso thanks for the PR! 

How is Switchboard being used within your proejct and if Switchboard oracles were hacked would your project lose >50% of its TVL? 
The reason we ask is that our goal with the dashboard is not just to list which oracles are being used by various projects, but to highlight how much TVL each oracle provider is securing. If an oracle is in use but doesn’t play a crucial role in securing the project, we wouldn’t want to include it. We’re specifically focused on cases where the oracle is vital to the protocol, and a hack or incorrect report would result in a loss of over 50% of the TVL

gin-meso:
Hey @realdealshaman, thanks for your question!

Some of our tokens get price from Switchboard, for example: https://app.switchboard.xyz/aptos/mainnet/feed/0x162322bc562e6f570ead94eacbabba5ff471ae8a6425de08c37cc57037e3224a.
We have verified the security when using Switchboard and our contract even if Switchboard is hacked it will not affect.

Thank you!

PR #8484, opened 2024-11-04
chore: add umbrella to oracles

0xsambugs:
We are now adding Umbrella Network to our list of oracles


realdealshaman:
Hey @0xsambugs where can we find this in your docs?

0xsambugs:
It isn't, but it was announced here https://twitter.com/UmbNetwork/status/1813944857838117362 and here https://medium.com/umbrella-network/umbrella-integrates-with-balmy-90cfa7f0247c

Is that enough?

PR #8507, opened 2024-11-08
Request to update LSP metadata

IrvinShek:


realdealshaman:
Hey @IrvinShek why change the url? Is the current one wrong? Where can we confirm the correct one?

IrvinShek:
Thank you. This updated is our official URL. The previous pointed to our dapp interface. We have already pushed the change in the community, such as Twitter. The introduction on Twitter or the button on the official website also points to previous. This is a necessary update.

PR #8552, opened 2024-11-14
fix: updated the description of a51 finance on parent and v2 tab

HamzaBhatti125:


HamzaBhatti125:
Hey @g1nt0ki could you please also update the description of a51 finance on the remaining places

PR #8635, opened 2024-11-25
Update chains in IPOR Derivatives and IPOR Fusion

rav-ipor:
Other updates in IPOR Derivatives and IPOR Fusion:
- url
- gecko_id
- cmcId
- wrongLiquidity


realdealshaman:
Hey @rav-ipor thanks for the PR! We only added the gecko id and cmc id to the parent protocol when one is present. Adding it to the individual listing triggers an error that helps us avoid having duplicate entries. I'll merge the PR and remove the duplicate entries. Thanks again

PR #8643, opened 2024-11-25
Update PV01 metadata for coingecko id and category

kevin-pv01:
For protocol PV01, we now have a coingecko listing.
Our category is better described as RWA. We are a yielding product, but use RWA to produce that yield, and our most close competitors are listing as RWA so we should follow suit.


realdealshaman:
Hey @kevin-pv01 thanks for the PR! We currently only use that space for governance token info. Does PV01 have a governance token?

kevin-pv01:
Ok @realdealshaman it is ready to review. I've removed the gecko_id and so now it is just a change of category.

PR #8664, opened 2024-11-26
kafka produce

waynebruce0x:


coins/src/utils/coins3/produce.ts
```diff
@@ -0,0 +1,76 @@
...
+async function produceTopics(
+  items: Dynamo[],
+  topic: Topic,
+  producer: Producer,
+) {
+  const messages: Message[] = [];
+
+  items.map((item) => {
+    const { symbol, decimals } = item;
+    if (!symbol || !decimals) return;
```
g1nt0ki, line 26: this would fail if decimals = 0

coins/src/utils/coins3/jsonValidation.ts
```diff
@@ -0,0 +1,61 @@
...
+  timeseries: {
+    type: "object",
+    properties: {
+      ts: { type: "integer" },
+      pid: { type: "string" },
+      price: { type: "number" },
+      confidence: { type: "number" },
+      source: { type: "string" },
+    },
+    required: ["ts", "pid", "price", "confidence", "source"],
```
g1nt0ki, line 44: confidence and source can be optional?

coins/src/utils/coins3/jsonValidation.ts
```diff
@@ -0,0 +1,61 @@
...
+    type: "object",
+    properties: {
+      decimals: { type: "integer" },
+      symbol: { type: "string" },
+      address: { type: "string" },
+      pid: { type: "string" },
+      chain: { type: "string" },
+      source: { type: "string" },
+    },
+    required: ["decimals", "symbol", "address", "pid", "chain"],
```
g1nt0ki, line 21: chain here can be optional

coins/src/utils/coins3/produce.ts
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  await producer.send({ topic: `${topic}`, messages });
+}
+function convertToMessage(item: Dynamo, topic: Topic): object {
+  const { PK, symbol, decimals, price, confidence, timestamp, adapter } = item;
+  const { chain, address, pid } = splitPk(PK);
+
+  switch (topic) {
+    case "metadata":
```
g1nt0ki, line 39: to futureproof, add 'coins-' as prefix to all three topics

coins/src/adapters/utils/database.ts
```diff
@@ -396,6 +397,7 @@ export async function batchWriteWithAlerts(
   const filteredItems: AWS.DynamoDB.DocumentClient.PutItemInputAttributeMap[] =
     await checkMovement(items, previousItems);
   await batchWrite(filteredItems, failOnError);
+  await produceKafkaTopics(filteredItems.filter((i) => i.SK == 0) as any[]);
```
g1nt0ki, line 400: this would mean, we write only current records? also, topic is missing here. Guessing this is sample code? 


So, other than minor comments, the code looks good, we should give it a try and write some to kafka, so we can test writing to our DB

waynebruce0x: the batch writes to dynamo are all duplicated (sk 0 and historical). Here I've filtered the items to sk0 only and then written them to all topics. This will write all topics.

coins/src/utils/coins3/jsonValidation.ts
waynebruce0x: I thought itd be a good idea to keep these as required. What would the use case be for undefined confidence or source?

coins/src/utils/coins3/produce.ts
```diff
@@ -23,7 +23,7 @@ async function produceTopics(
 
   items.map((item) => {
     const { symbol, decimals } = item;
-    if (!symbol || !decimals) return;
+    if (!symbol || decimals == null) return;
```
g1nt0ki: typeof decimals !== 'number' is safer? or it will skip decimals which are string?  hmm... maybe we can leave this as is
image

coins/src/utils/coins3/produce.ts
```diff
@@ -69,6 +69,7 @@ export default async function produce(
   items: Dynamo[],
   topics: string[] = allTopics,
 ) {
+  if (!items.length) return
```
g1nt0ki: add `;` in the end to be safe

coins/src/utils/coins3/jsonValidation.ts
g1nt0ki: souce is not strictly necessary and missing in lot of defi adapters right? my fear is, we silently reject some tokens and wont notice it

coins/src/adapters/utils/database.ts
g1nt0ki: Hmm, I would rather split between 0 & with timestamp, write the timestamp ones too time series & without ts to current, I guess the mapping for both needs to be sent to mapping topic?

g1nt0ki: My worry is, what happens when you refill coin data, then it won't be sent to the topics right? Like filling nft prices a month ago

PR #8667, opened 2024-11-26
Add chainlink oracle to APX Finance

JoaoVasques:
Updating the list of oracles to include Chainlink. Proof is in the link that was previously in the code.

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -9138,7 +9138,7 @@ const data2: Protocol[] = [
     cmcId: "16334",
     category: "Derivatives",
     chains: ["Binance"],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @JoaoVasques thanks for the RP! I was looking through the docs and Chainlink appears to function as a backup oracle in APX, as it is used to verify the accuracy of Pyth Oracle's prices and trigger a circuit breaker if discrepancies exceed 1%. In these cases we only add the Primary oracle

JoaoVasques:
> Hey @JoaoVasques thanks for the RP! I was looking through the docs and Chainlink appears to function as a backup oracle in APX, as it is used to verify the accuracy of Pyth Oracle's prices and trigger a circuit breaker if discrepancies exceed 1%. In these cases we only add the Primary oracle

Hi @realdealshaman! Thanks for the clarification, I will close the PR then.

PR #8708, opened 2024-12-02
Change sacra image and remove parent protocol

alexandersazonof:


defi/src/protocols/data3.ts
```diff
@@ -62195,7 +62195,6 @@ const data3: Protocol[] = [
     twitter: "sacra_fi",
     forkedFrom: [],
     oracles: [],
-    parentProtocol: "parent#tetu",
```
Define101, line 62198: Why remove this?

alexandersazonof: Because its different projects

alexandersazonof: @Define101 I can return if its important

realdealshaman: @alexandersazonof we also group projects together when they are part of the same team. I notice that they are all being worked on from the tetu github repo

alexandersazonof: return parent protocol

alexandersazonof: @realdealshaman ping

PR #8746, opened 2024-12-05
Update Upshift Category

krishnafractol:


realdealshaman:
Hey @krishnafractol considering that the strategy is for users to lock tokens in a vault that will then be managed by a strategist. I believe that the best category is Managed Token Pools.

'Managed Token Pools':	'Protocols where token pools are actively controlled and managed by a designated operator or governance',

PR #8790, opened 2024-12-11
Add-pyth-new-protocols

mariobern:
gm llama sirs

adding Satori that has integrated with the Pyth oracle

https://docs.satori.finance/perpetual-trading/index-price

thank you


mariobern:
image

PR #8811, opened 2024-12-13
Fraxtal:Fraxlend - Oracle Provider Update 12/13/2024

tom2o17:


tom2o17:
Hey @realdealshaman, 

I can add it to the frax docs if you would like. 

Otherwise the pairs are visible through the UI [here](https://app.frax.finance/fraxlend/available-pairs): 
Screenshot 2024-12-16 at 11 41 23 AM

And on the contracts themselves [here](https://fraxscan.com/address/0x1174B74aCB27CB20F2ACD12a324deD7a1ce16041#readContract): 
Screenshot 2024-12-16 at 11 41 48 AM

realdealshaman:
@tom2o17 thanks for the additional info :pray: , from the market data, it seems that these two are very small still. Not enough deposits in them to attribute the Security of the entire TVL to API3. If these markets are later 50% of the TVL or if API3 is used for others and end up providing data for the majority of the TVL we can update the "oracle used for securing" info

PR #8816, opened 2024-12-15
update bluefin oracle

aka-poke:
saw an update on Bluefin and take it to myself to create a PR for Bluefin oracle 

https://learn.bluefin.io/bluefin/bluefin-exchange/trading/oracle-and-index-price#oracle-price


realdealshaman:
Hey @aka-poke thanks for the PR! I was able to look through the docs and it seems that it is referring to the Bluefin perps product which is listed at id: 2625 on data2.ts and not the AMM

realdealshaman:
no problem, Thanks for the PR! :pray:

PR #8860, opened 2024-12-18
swap price update

waynebruce0x:


coins/src/updateCoin.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (PK in bulk && bulk[PK] > unixStart - margin / 2) return;
+    writes.push(
+      ...[
+        {
+          PK: `coingecko#${id}`,
+          SK: 0,
+          price,
+          mcap,
+          timestamp: SK,
```
0xngmi, line 63: maybe we also store current timestamp as a new field? that way if a coin is outdated we wont keep requesting updates

waynebruce0x: added a case for this in next commit

PR #8866, opened 2024-12-19
Update Gearbox category, description and audits link.

de-snake:
The new category is on the website, in materials, in new vision, and interface. Description is changed accordingly
- landing proof: https://gearbox.fi/
- interface app proof: https://app.gearbox.fi/lending


realdealshaman:
Hey @de-snake thanks for the PR! Will this "Lending" be a new product? If deposits will be held separately, then it is better to create a new adapter for the Gearbox Lending and we can also do a new listing

de-snake:
Hey. No, lending is the main product and it is already being utilized as a lending protocol
The 1-click leverage interface may be more popular for now, but that's only a frontend ux feature that simplifies interaction with a core lending protocol

Comparing to other protocols:
- Contango is a pure FE, has no own liquidity sources, hard to use for simple borrowing
- Alpaca/Extra finance etc are pure leveraged farms, have its own liquidity but no option to simply borrow because of the single vault-based architecture (afaik)
- Aave is a pure lending protocol: own liquidity, but without UX optimizers atop of it it's hard to loop to a high leverage

Gearbox has its own liquidity, architecture is flexible for all the mentioned usecases (i.e. we have built-in multicalls & router to abstract swapping, staking, redeeming and other actions that can be needed to create and unwrap a complex defi positions involving borrowed tokens)

de-snake:
@realdealshaman pinging for visibility

PR #8879, opened 2024-12-19
Update Javsphere Platform infos and token

rogi-sh:
Project update and add important infos


realdealshaman:
Hey @rogi-sh thanks for the PR! There's no need to add mapping for your token, we already price it. Please remove those changes and we can merge your PR

rogi-sh:
@realdealshaman done. Just wondered why the token price is not on the main page.

PR #8889, opened 2024-12-20
added Pyth oracle for Echo Lending

echo-pro-dev:
added Pyth oracle for Echo Lending

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -53377,7 +53377,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Lending",
     chains: ["Aptos"],
-    oracles: [],
+    oracles: ["Pyth"],
     forkedFrom: [],
     module: "echo-lending/index.js",
     twitter: "EchoProtocol_",


realdealshaman:
Hey @echo-pro-dev thanks for the PR! Are there some docs that we can reference for info on the way the oracle is being used?

echo-pro-dev:
Currently, all assets on Echo Lending use Pyth Oracle prices.
Usage scenarios:
  1. After the user deposits asset A, the total value of asset A can be calculated based on the price, and then the maximum borrowable value can be calculated based on the LTV.
  2. When the user borrows asset B, the maximum borrowable amount will be calculated based on the maximum borrowable value and the price of B.
  3. When the user needs to be liquidated, the latest price must be updated before the liquidation operation can be performed.

Here are the related codes:
Echo Oracle based on Pyth, https://explorer.aptoslabs.com/account/0xf8e9225f460f75e2ffde5c6da5cd99aea2a3d14e93b18327a43a11c1c57c5ae0/modules/run/oracle/add_asset?network=mainnet

Liquidation interfaces, https://explorer.aptoslabs.com/account/0xeab7ea4d635b6b6add79d5045c4a45d8148d88287b1cfa1c3b6a4b56f46839ed/modules/run/router/liquidation_coin?network=mainnet

realdealshaman:
Thanks for the additional info and links :+1:

PR #8943, opened 2024-12-27
add pyth to tren + neutral trade

mariobern:
gm llama sirs

adding Tren that uses both Pyth and Chainlink, as well as Neutral Trade that uses Pyth

Tren
https://docs.tren.finance/protocol/asset-risk/oracle-risk

Neutral Trade
https://docs.neutral.trade/additional-info/pyth-oracles

thank you


realdealshaman:
@mariobern thanks for the PR!

PR #8954, opened 2024-12-30
>100 new gecko id's and utility functions

SecureZer0:
i added >100 new gecko_ids for tokens by matching the twitter handle and or the homepage url to that on coingeckos /coins endpoint. 

I added some filtering to reduce potential errors like adding stablecoins or restaked token inplace of the desired equity-like tokens and not entering tokens without a current price

I then went through the list of new entries and checked for any incorrect coingecko_ids, i still believe its worth creating tests that checks to see if any gecko_ids are incorrect, for both this data and existing data in defillama

The utility function to insert coingecko_ids to the data1/2/3.ts & parentProtocol.ts files are in defi/src/cli/


0xpeluche:
#8969

PR #8978, opened 2025-01-03
add switchboard oracle to Carrot protocol 

jsblair9:
Updated the Carrot protocol to list Switchboard as an oracle.

It is specific to Solana.

The oracle feed is here: https://ondemand.switchboard.xyz/solana/mainnet/feed/GQbKvThYgyvdp8T8MoE8KKD2NngXumQ8mDkCLxmhvJKm

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -59738,7 +59738,7 @@ const data3: Protocol[] = [
     module: "carrot/index.js",
     twitter: "DeFiCarrot",
     forkedFrom: [],
-    oracles: [],
+    oracles: ["Switchboard"],
     audit_links: ["https://deficarrot.com/assets/carrot_final_report.pdf"],
     listedAt: 1731107322
   },
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -59738,7 +59738,7 @@ const data3: Protocol[] = [
     module: "carrot/index.js",
     twitter: "DeFiCarrot",
     forkedFrom: [],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @jsblair9 thanks for the PR! I am able to see in your link that switchboard has a price feed for CRT but how do we know that it is being used within their protocol?

realdealshaman:
Is that info documented somewhere?

realdealshaman:
I can only find that Pyth is being used:
image

jackrieck:
Thanks for the comments. pyth oracles are used for the treasury of assets like (USDC/USDT/pyUSD), this is what is discussed in the audit, but the protocol writes the value of CRT into a switchboard account. Here it is https://ondemand.switchboard.xyz/solana/mainnet/feed/GQbKvThYgyvdp8T8MoE8KKD2NngXumQ8mDkCLxmhvJKm

realdealshaman:
Hey @jackrieck thanks for the feedback. It seems in this case both oracles are important since the stablecoins backing CRT are priced by Pyth but the price of CRT is determined by Switchboard which could affect redemptions. Is this correct?

jsblair9:
Hey @realdealshaman -- I'm unsure what the ask is here. Happy to provide more context if needed!

would you like for us to include Pyth on our list?

jackrieck:
> Hey @jackrieck thanks for the feedback. It seems in this case both oracles are important since the stablecoins backing CRT are priced by Pyth but the price of CRT is determined by Switchboard which could affect redemptions. Is this correct?

Yup that's correct, what should we put?

PR #8990, opened 2025-01-06
more tokens

waynebruce0x:


0xqiba:
Hi, I think there may be a mistake here -- the FUSDC token on coingecko with id "fluid-usdc" is not the same as the Fluid fUSDC tokens on eth/arb.

PR #9035, opened 2025-01-14
Update category of NODO

cuongnp-ibl:


realdealshaman:
Hey @cuongnp-ibl based on NODO's core focus on short-term price forecasts (0DTE) and real-world event prediction polls, where users wager on outcomes for rewards - "Prediction Market" is the best fit. It directly aligns with protocols that let users bet on future results and earn payouts based on accuracy

PR #9054, opened 2025-01-15
Update Biswap Token data and description

ivicmad:


realdealshaman:
Hey @ivicmad thanks for the PR! However, we can't have a child protocol listed with the same name as the parent protocol

ivicmad:
Please check new naming

realdealshaman:
Hey @ivicmad thanks! I'll merge it but will still need to make further corrections. The coingecko ID and CMC ID are only needed on the parent protocol listing. Thanks again for the PR

PR #9087, opened 2025-01-20
swell description update and add cmc, gecko ids

darvinrio:


realdealshaman:
Hey @darvinrio thanks for the PR! Actually, only the parent protocol entry should have the coingecko and cmc ids. To avoid duplicates. I've gone ahead and updated the description as requested in the PR. Thanks again

PR #9143, opened 2025-01-24
update craftChainsResponse to include defillama_id in it's response

ferenci84:
Endpoint /v2/chains do not include an id that can be used for the /prices/current/{coins} endpoint as the chain id for the {chain}:{address} pattern. This PR fills this gap adding defillama_id.

### DIFF
--- a/defi/src/getChains.ts
+++ b/defi/src/getChains.ts
@@ -6,12 +6,12 @@ import { IChain, } from "./types";
 import { importAdapter } from "./utils/imports/importAdapter";
 import { excludeProtocolInCharts, isExcludedFromChainTvl } from "./utils/excludeProtocols";
 
-async function _checkModuleDoubleCounted(protocol: Protocol){
+async function _checkModuleDoubleCounted(protocol: Protocol) {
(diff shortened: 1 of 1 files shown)


defi/src/getChains.ts
```diff
@@ -6,54 +6,61 @@ import { IChain, } from "./types";
...
+        }
+        chainTvls[chainName].tvl += chainTvl as number
+        chainsAdded += 1;
       })
-      if(chainsAdded === 0){ // for fetch adapters
+      if (chainsAdded === 0) { // for fetch adapters
         const chainName = protocol.chain
-        chainTvls[chainName] = (chainTvls[chainName] ?? 0) + lastTvl.tvl
+        if (!chainTvls[chainName]) {
+          chainTvls[chainName] = { tvl: 0, defillama_id: chainName }
```
0xngmi, line 50: this is wrong because this will show the formatted chain name instead of the raw defillama id

0xngmi (changes requested):

ferenci84:
@0xngmi Can you please review again?

PR #9177, opened 2025-01-29
Updated Crypto Valley Exchange (Description, Name, UCID)

siomochkin:
Hello,
We would like to update the description of our project to align with our current positioning and strategy. Additionally, we want to add “(CVEX)” to the name to enhance its visibility in search results. We have also discovered the possibility of including a cmcid, which means we won’t need to make any further changes after the TGE.
Please let me know if you have any questions.
Thank you!


siomochkin:
The previous PR was submitted by my colleague https://github.com/vladjito, please tag him if you need any confirmation.

realdealshaman:
Hey, @siomochkin thanks for the PR! One reason that we don't apply this type of name is that it is too long and does not display correctly, also the parenthesis don't allow the logo to appear correctly

siomochkin:
@realdealshaman thanks for the feedback, reverted name change

PR #9191, opened 2025-01-31
fix: update levvy for nfts category to Lending

ricomiles:


realdealshaman:
Hey @ricomiles thanks for the PR but if the product is NFT lending protocol then the current category is accurate

ricomiles:
Hello, @realdealshaman We would like to generalize this to “Lending,” so that the parent protocol, “Levvy-Finance,” will display “Lending” instead of an empty category.

realdealshaman:
@ricomiles it shows empty because there is more than one product, even if both are "Lending" since there are two separate products and listing it will still show as is. You need to click the arrow next to the name to open the drop down and show the categories
image

ricomiles:
@realdealshaman oh I see that make sense. With that in mind, is there any way we can have the parent protocol show a category? Thank you so much for your assistance.

realdealshaman:
@ricomiles I get the idea, but it’s not something we’re planning to add since it wouldn’t make sense in most cases with multiple products.

PR #9192, opened 2025-01-31
Suilend: integration with Switchboard

mgild:
This PR reflects Suilend switchboard integration


realdealshaman:
Hey @mgild thanks for the PR! Please also include info on how it is used, for what assets and how much of their total tvl those assests represent. Thanks

realdealshaman:
the frontend shows Pyth is providing pricefeeds for all the assets

PR #9193, opened 2025-01-31
Switchboard jito integration

mgild:
Add Switchboard dependency for Jito.

Jito has just launched their tip router which permits multi-token tipping to their platform.

Switchboard is used as the oracle to price all collateral.

https://github.com/search?q=repo%3Ajito-foundation%2Fjito-tip-router%20switchboard&type=code

https://x.com/switchboardxyz/status/1885342385242333228


mgild:
The tip router is Jito's new product to price all collateral for jito's priority fees (was 2 Billion dollars for 2024)

These MEV tips feed directly into the jitoSOL yield rewards

realdealshaman:
@mgild Does this introduce any risk to the SOL locked within Jito if incorrect data is provided?

mgild:
@realdealshaman

To explain a bit here on background: https://www.jito.network/blog/what-is-jito-tiprouter/

To explain in terms of the support around the JitoSOL LST and how this plays into jito tips,

JitoSOL LST increases staker yield by acting as both an LST and a MEV tip distributor.

As you can see here: https://dune.com/queries/3970023/6680731, almost half of jitosol yield is now encompassed by their MEV tips.

As Switchboard now prices the tip router and is securing Jito tip revenue, this directly correlates to the user yield and sticking power to JitoSOL as well as directly attached to Jito's yearly MEV revenue

realdealshaman:
Hey @mgild, thanks for the explanation! I understand this product is important for attractiveness of Jito’s staking, but it doesn’t secure the staked SOL itself. Even if users withdrew due to dissatisfaction with yield, the oracle wouldn’t be directly securing the TVL. So, it doesn't qualify for our Total Value Secured Dashboard

mgild:
Hi @realdealshaman thank you for the follow up. Securing all jito rewards is a huge amount of volume that not being reflected on defillama dashboard would misrepresent switchboard. What would you suggest we do to have that reflected accurately?

realdealshaman:
Hey @mgild I've added oracleBreakdown info to Jito to note that Switchboard is being used there:
image

mgild:
Thank you @realdealshaman, what does secondary denote?

mgild:
Also how will this show up on the oracles page?

https://defillama.com/oracles/Switchboard

realdealshaman:
@mgild 

realdealshaman:
it has not been integrated to the UI at this time, there are still many entries missing but we plan on eventually having this data showed as well

PR #9196, opened 2025-02-01
Angles yield

anglesb:
Hey. Submitting yield tracker adapter for our LST project on Sonic (https://defillama.com/protocol/angles) based on ERC4626 value growing strategy with every rebase


waynebruce0x:
https://github.com/DefiLlama/defillama-server/pull/9230
This PR achieves the same. We have no price for the underlying, so wanS will still not be priced. Please let us know in discord how to price anS..

anglesb:
> #9230 This PR achieves the same. We have no price for the underlying, so wanS will still not be priced. Please let us know in discord how to price anS..

anS is issued 1:1 with S so it has the same price as S (Sonic)

PR #9232, opened 2025-02-05
Update protocol oracle mapping

JoaoVasques:


realdealshaman:
Hey @JoaoVasques thanks for the PR!

realdealshaman:
I removed the entry for UntitledBank. I see that there are Chainlink price feeds but on their UI, we can see that the majority of the TVL is being priced by Pyth

PR #9239, opened 2025-02-05
Update Primex Finance info

Mar-Alex:


realdealshaman:
Hey @Mar-Alex, thanks for the PR!

The first chain key indicates where TVL was first listed - please leave it as chain: "Polygon".
For oracles, could you specify which assets each oracle provides price feeds for? We should only keep the primary oracle.

Looking forward to merging this. Thanks again

PR #9277, opened 2025-02-10
Update Xswap prtocol data on Crossfi chain

0xMarkuss:
Updated xSwap metadata

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -65102,7 +65102,7 @@ const data3: Protocol[] = [
     description:
       "Exchange tokens issued on the CrossFi Chain",
     chain: "CrossFi",
-    logo: `${baseIconsUrl}/crossfi-xswap.jpg`,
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey, @0xMarkuss where can we find some proof of these changes? Is there any public announcement?

0xMarkuss:
Hey @realdealshaman, here is our media 
https://x.com/xswapms
https://www.facebook.com/share/15dgu6LqrE/?mibextid=wwXIfr

and link to our protocol's app
https://xswap.ms

and also you could find partner reposts in official CrossFi chain media (partnership announce was few month ago)

https://x.com/crossfichain



also this changes caused due to not enough data was provided on inotial xSwap adjustment cause main goal was to add CrossFi chain 

so now i want to provide data to actualise xSwap's info


regarding the protocol naming here was a little issue: originally the name is xSwap but it is already taken 
so we 're decided  to make is xSwap DEX to avoid misunderstanding from our users side

PR #9280, opened 2025-02-10
add pyth oracle to protocols

mariobern:
gm

adding new pyth protocols:

MachFi:
Pyth is the primary oracle for MachFi (on top of API3), as noted in the audit results https://machfi.gitbook.io/machfi-documentation/contracts/audit

Neptune:
Pyth is the primary oracle for Neptune https://docs.hydrogenlabs.xyz/neptune-protocol/protocol-design/security

WaveX:
Pyth is used with Chainlink as the main oracles for the perp protocol https://wavex.fi/

Beraborrow:
Pyth is the main oracle for Beraborrow, as noted in the audit results https://beraborrow.gitbook.io/docs/audits/audits

Quill Finance:
Pyth is used with Chainlink as the main oracles for the stablecoin protocol https://docs.quill.finance/faq/liquidations-oracle-and-redemptions

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -65043,7 +65043,22 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Lending",
     chains: ["Sonic"],
-    oracles: ["API3","Pyth"], // https://machfi.gitbook.io/machfi-documentation/contracts/audit
+    oracles: ["API3"], // https://machfi.gitbook.io/machfi-documentation/contracts/audit
+    oraclesBreakdown: [
+      {
+        name: "API3",
+        type: "Primary",
+        proof: [
+          "https://machfi.gitbook.io/machfi-documentation/protocol/mechanics/risk-management#oracle-risk"
+        ],
+      },
+      {
+        name: "Pyth",
+        type: "Secondary",
+        proof: [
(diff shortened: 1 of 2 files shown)


realdealshaman:
Hey @mariobern thanks for the PR! 

The Quill Finance page does not appear ready, gives a page does not exist error:
image

also the Halborn audit of Beraborrow says that they will be using Redstone
image

Neptune appears to be using Switchboard as one of the main focuses of the audit is the recency check for the Switchboard oracles

MachFi uses pyth as a complementary, secondary oracle, to API3:
image

For waveX unfortunately I am not finding any info on how they are using the two oracle integrations but we are working on adding all oracles used so that we can present all of this info and all the integrations are visible. You can see an example of how we are beginning to add this with what I added to the MachFi listing (oracleBreakdown)

These are all of the different bits of data that we will be able to add:
oraclesBreakdown?: Array<{
    name: string,
    type: "Primary" // Oracle that secures more than 50% of protocol TVL, if oracle is hacked >50% of TVL will be lost
      // pls add more as needed
    proof: Array<string>,
    startDate?: string, // YYYY-MM-DD
    endDate?: string,
    chains?: Array<{
      chain: string,
      startDate?: string,
      endDate?: string
    }>
  }>

realdealshaman:
Hey, @mariobern I reviewed the docs in more depth and have added Pyth as the oracle to : Neptune Protocol, waveX and Beraborrow

Thanks again for the PR!

PR #9284, opened 2025-02-11
Add switchboard oracle to Fragmetric and fix chain to Solana

Kangji:


realdealshaman:
Hey @Kangji are there any docs that I can reference about how the oracle is being used?

realdealshaman:
Thanks for also flagging the chain, I've updated the listing to "Solana"

Kangji:
Hi, @realdealshaman. We're using normalized SOL (nSOL) for restaking at Jito. Jito restaking vault utilizes our nSOL price feed from switchboard. Please see https://ondemand.switchboard.xyz/solana/mainnet/feed/2s93wJ6CpYNurLzzbDDZGVsC6xPuvRjeqaTUG6gxJUZb

realdealshaman:
@Kangji Does Fragmetric’s TVL fully rely on Switchboard’s nSOL price feed, and could incorrect data from the oracle put its TVL at risk?

Kangji:
@realdealshaman TVL is determined by the underlying asset value of nSOL. You can find the vault associated with these assets in our directory. We do not rely on an oracle for our internal calculations—only as a fallback. However, Jito Restaking uses Switchboard’s oracle data for nSOL to evaluate our vault, and we maintain this integration to ensure the accurate valuation of our restaking vault.

realdealshaman:
@Kangji Got it! Since incorrect data from the oracle does not affect Fragmetric’s TVL, we won’t attribute its TVL to Switchboard’s oracle. However, we can note its integration as follows:
    oraclesBreakdown: [
      {
        name: "Switchboard",
        type: "Fallback",
        proof: ["https://github.com/DefiLlama/defillama-server/pull/9284"]
      }
    ],

realdealshaman:
these are the type defenitions:

realdealshaman:
I've added the entry here:
image

PR #9296, opened 2025-02-11
Add RedStone to oracles used by Resolv

ChmilevFA:
Requesting to add RedStone as an oracle provider for Resolv next to Pyth. 
RedStone and Pyth secure Resolv's collateral pool operations in terms of the price estimations. Incorrect oracle data from RedStone could expose the protocol logic to an exploit while managing inventory, hedging collateral and serving customer's requests.
https://docs.resolv.xyz/litepaper/protocol-mechanics/collateral-pool


realdealshaman:
Hey, @ChmilevFA thanks for the PR! Can you tell us which assets each oracle is pricing?

ChmilevFA:
> Hey, @ChmilevFA thanks for the PR! Can you tell us which assets each oracle is pricing?

Both RedStone and Pyth are utilized across many different assets in our system to ensure redundancy and avoid a single point of failure.

realdealshaman:
@ChmilevFA thank you for clarifying. Are there other oracles used for some of the assets?

realdealshaman:
and can the collateral pool also be drained by manipulating the USR or RLP price on redemptions? I see Chronicle is mentioned as a price oracle for USR

ChmilevFA:
> @ChmilevFA thank you for clarifying. Are there other oracles used for some of the assets?

Not at the moment. We might be adding new ones potentially, but for such a case we'll create a separate PR, the PR is only about adding RedStone to the list

ChmilevFA:
> and can the collateral pool also be drained by manipulating the USR or RLP price on redemptions? I see Chronicle is mentioned as a price oracle for USR

on direct burns/redemptions in regards of USR or RLP market prices - no. For that we are using our own "fundamental" price to USD.

realdealshaman:
@ChmilevFA we've combined the USR and RLP listings previously into the single Resolv listing so we only need to add the oracle data there. I've added it in the following commit: https://github.com/DefiLlama/defillama-server/commit/f17e40ff3f9bda133dfe41576c14a9774d3d2fa7

Thanks again for the help

PR #9328, opened 2025-02-18
Update BetSwirl protocol

RomualdH:
Updating the chains and audits link.


realdealshaman:
Hey @RomualdH the first chain entry notes the first chain that the project was on when it was initially listed, we can't update that portion beyond the initial listing date: chain: "Binance",

RomualdH:
Hey @realdealshaman, I've reverted this change.

PR #9344, opened 2025-02-21
Update data3.ts

jan-magpie:
Hello, 
requesting to add RedStone as an oracle provider for EigenPie. RedStone secures our staking-oriented product, and helps us calculate yield of supported strategies. Incorrect oracle data from RedStone could expose strategies to significant risks through inaccurate valuations and yield calculations, potentially leading to losses, a drain of the vault or inefficient capital allocation. Since RedStone provides critical data for yield calculations, its reliability directly impacts the EigenPie’s system stability.

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -31469,7 +31469,7 @@ const data3: Protocol[] = [
     cmcId: null,
     category: "Liquid Restaking",
     chains: ["Ethereum"],
-    oracles: ["Chainlink"],
+    oracles: ["Chainlink","RedStone"],
     forkedFrom: [],
     module: "eigenpie/index.js",
     twitter: "Eigenpiexyz_io",


realdealshaman:
Hey, @jan-magpie thanks for the PR, we've received similar PRs but this seems that RedStone is not directly securing the TVL instead it could lead to losses from users removing their own deposit but not from a hack. We can add it but as a Fallback oracle

jan-magpie:
Thanks @realdealshaman for the feedback. RedStone is used at par with Chainlink in the system.
It is true that RedStone’s mispricing would lead to losses from users removing their own deposit, but also those adding new deposits. As a result, an attacker could take advantage of incorrect ratios and drain deposits. Hence, RedStone should be added next to Chainlink.

realdealshaman:
@jan-magpie I understand the logic but we focus on whether a hacker could steal funds by manipulating the oracle, since the answer is no, then we can say it is securing the TVL

PR #9345, opened 2025-02-21
added oracle for Neemo Finance

Alfa11917676:


realdealshaman:
Hey @Alfa11917676 thanks for the PR! Can you tell us a bit more about how Chainsight is being used within Neemo? we'd also need some docs that we can reference please

Alfa11917676:
Hey @realdealshaman ,
Thanks for reviewing our PR! We are using Chainsight to fetch the current price from DEXs, which is also utilized by Untitled Bank for leverage trading. Below is a code snippet where we’ve integrated it into our codebase. Let me know if you have any questions or suggestions.
image

realdealshaman:
@Alfa11917676 Thanks for the snippet but it still doesn't show code that shows a user can deposit X ASTR and somehow exploit a misprice to mint excessive nsASTR though, this snippet just shows that a price feed(via Chainsight) is provided for others to use, Untitled Bank in this case

Alfa11917676:
@realdealshaman , I didn't get your question. As, using _convertToken() we are minting tokens to users, and the chainsight oracle is only used for untitled bank for their leverage trading requirement.

realdealshaman:
Hi @Alfa11917676 can you please explain how Chainsight is securing the TVL though? If Chainsight's oracle were hacked would Neemo funds be at risk and how?

Alfa11917676:
Chainsight plays a crucial role in providing the real-time market rate of nsASTR, which is essential for external vendors like Untitled Banks to access accurate trading prices from DEXs. While Chainsight enhances transparency and ecosystem growth, the minting price of nsASTR is determined solely by our internal `_convertToken()` function. This ensures that even if Chainsight were compromised, the minting mechanism and Neemo’s TVL would remain secure. Chainsight’s oracle contributes significantly to the usability and adoption of nsASTR by enabling reliable price discovery for external integrations.

realdealshaman:
Hey @Alfa11917676 sorry, that is not the intended use for the oracle dashboard, we only add oracles to project when there is risk that the protocol will have the user funds stolen if the oracle gets hacked

PR #9352, opened 2025-02-22
Update token information for Amped

AmpedFinance:


realdealshaman:
Hey @AmpedFinance thanks for the PR! However there are some things that we can't change from the listing info, the first "chain" can't be udpated since it notes the first chain that was added when we originally did the first listing and also the listedAt entry can't be updated since it is used to note the date of the original listing

AmpedFinance:
Ok, thanks @realdealshaman 

We'll leave it as is

PR #9355, opened 2025-02-24
Add SolvBTC and SolvBTC.BBN in soneium network

buchaoqun:


Define101:
We already price those

PR #9369, opened 2025-02-26
Update category for Elixir

cairoeth:


realdealshaman:
Hey @cairoeth thanks for the PR! In this case RWA is not the best category since the reason users are interacting with Elixir is for earning yield and leveraging RWAs in DeFi, while tokenizing RWAs isn’t the direct user action

PR #9374, opened 2025-02-26
Update Oracles for JonesDAO.ts

Hundert1000:
Hey Llamas,
Kindly asking to update oracles for JonesDAO

Oracle Provider(s): RedStone

Implementation Details: JonesDAO is using RedStone as the primary oracle on main pools.

Documentation/Proof:
https://docs.jonesdao.io/jones-dao/other/security#oracles

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -1280,7 +1280,7 @@ const data2: Protocol[] = [
     cmcId: "17743",
     category: "Yield Aggregator",
     chains: ["Arbitrum"],
-    oracles: [],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @Hundert1000 thanks for the PR! However, we'll need a clear explanation of how you determine that RedStone is securing the TVL of Jones DAO. Jones DAO’s TVL includes a large amount of fsGLP - about $3M out of $5.4M total. Because fsGLP’s value is derived from GLP’s underlying basket of assets (which use Chainlink feeds), that $3M portion relies on Chainlink oracles. RedStone is not securing those Chainlink-dependent values. So we can’t say RedStone secures the entire Jones DAO TVL, since a majority of it depends on Chainlink pricing

PR #9378, opened 2025-02-27
Feat: submit update sVIC

khiemsoft:


realdealshaman:
Hey @khiemsoft thanks for the PR! However we only add coingecko and cmc ids when the project has a governance token
same for the address and ticker

PR #9388, opened 2025-02-28
Update Oracle for Silo V2 on Sonic.ts

Hundert1000:
Hey Llamas,
Kindly asking to update oracles for Silo on Sonic.

Oracle Provider(s): RedStone

Implementation Details: Sonic is using RedStone on it's biggest S/stS market, and a few other markets. (Can be checked [here](https://v2.silo.finance/).

Zrzut ekranu 2025-02-28 o 18 33 36

Overall RedStone feeds are securing ~$178m across markets which is around 46,5% of Silo sonic TVL. Since the impact of the hack would be significant here, kindly asking to add RedStone to Silo reflected oracle setup.

### DIFF
--- a/defi/src/protocols/data4.ts
+++ b/defi/src/protocols/data4.ts
@@ -763,7 +763,7 @@ const data4: Protocol[] = [
     category: "Lending",
     chains: ["Arbitrum","Sonic"],
     oraclesByChain: {
-      sonic: ["Chainlink"], // https://silopedia.silo.finance/oracles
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @Hundert1000 in this case we still consider it a secondary oracle based on the following options: 
  oraclesBreakdown?: Array<{
    name: string,
    type: 

realdealshaman:
I'll add it as a secondary oracle under oraclesBreakdown, when it is above 50% give me a ping and I'll update it to Primary

PR #9418, opened 2025-03-06
update Bit Reserve category to CDP

bit-reserve:
update bit reserve category from Liquid Staking to CDP 

Users deposit liquid staking tokens (LSTs) into the protocol to mint RS tokens, representing collateralized debt positions (CDP). These RS tokens are non-liquid staking positions , allowing holders to earn continuous yield from underlying deposited assets

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -44089,7 +44089,7 @@ const data3: Protocol[] = [
     audit_note: null,
     gecko_id: null,
     cmcId: null,
-    category: "Liquid Staking",
+    category: "CDP",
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @bit-reserve, thanks for the PR! but unless Bit Reserve allows users to deposit collateral and mint a stablecoin (like MakerDAO’s DAI), it does not fit under CDP. Instead, it belongs in "Yield Aggregator" based on how it actually works: collects and optimizes yield from several sources, compounds earnings to optimize the returns for RS holders, and it integrates with several defi projects to maximize capital efficiency

PR #9434, opened 2025-03-08
chore: update avalon USDa and sUSDa

jasmin-avalon:


waynebruce0x:
I believe the coingecko ID of avalon USDa is `usda-2` rather that `usda`
Screenshot

jasmin-avalon:
Yes. You're correct. I have updated the coingecko ID of avalon USDa to `usda-2`

PR #9437, opened 2025-03-10
update Desyn Primary oracles

santinoYoung:


santinoYoung:
Hi, team. I flushed master to pull down the code and re-updated the desyn oracles config, but why does it show a conflict？

realdealshaman:
Hey @santinoYoung, thanks for the PR! I looked into DeSyn's products, and it seems that DeSyn Safe is the least likely to need oracles, as it involves simple staking. Could you clarify how the oracle is securing the TVL? I'll remove Chainlink from that listing for now, as it may have been added earlier based on how the project's TVL was tracked, before the products were clearly separated. Looking forward to your response to better understand how the oracle is used and to ensure it's attributed to the correct product. Thanks again!

santinoYoung:
@realdealshaman HI, TEAM, thanks for the reply. We have three types of strategies DeSyn Safe, DeSyn Basis Trading, DeSyn Liquid Strategy, for DeSyn Basis Trading, DeSyn Liquid Strategy strategies, we are using off-chain stats tvl, and off-chain oracle comes from Redstone , chainklink and TWAP and CEX, so I moved the file from under desyn safe to DeSyn Basis Trading.

PR #9468, opened 2025-03-14
update Allbridge listing with new chains list

m-cote:


realdealshaman:
Hey @m-cote thanks for the PR! The following chains are not yet added to the the Allbridge Classic adapter, we can't add these or it will cause errors: "Stacks", "Klaytn", "Stellar"

realdealshaman:
I'll merge this and remove the missing chains from the listing. Thanks again

PR #9476, opened 2025-03-17
Update level description

davidlee1435:


realdealshaman:
Hi, @davidlee1435 thanks for the PR! We should not update the category here since a CDP is primarily about minting a debt-backed asset by locking collateral, while a yield-bearing stablecoin is about earning passive income from deposited stablecoins

davidlee1435:
hey! apologies for the late reply. we're looking to make the change for a couple reasons:

1. we were first categorized as a CDP by the defillama team. other USDC/T backed stablecoins also seem to have this distinction (ie see Sky Protocol, Reservoir Protocol, DSU Money)
2. the largest protocol in the yield section are Pendle, Royco and Convex, which are very different to us

i think the best category would be to create a stablecoin category for all of [the protocols here](https://defillama.com/stablecoins)- any thoughts there?

PR #9509, opened 2025-03-21
Update of Zest Oracle

bats4:
Hi Llamas!

another PR for update on Zest Oracle

Docs
https://docs.zestprotocol.com/start/borrow/zest-protocol-borrow-overview/oracles#zest-protocol-uses-pyth

X post
https://x.com/ZestProtocol/status/1902782204121686174

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -39466,6 +39466,14 @@ const data3: Protocol[] = [
     twitter: "ZestProtocol",
     forkedFrom: [],
     oracles: ["Pyth"], // https://docs.zestprotocol.com/start/borrow/zest-protocol-borrow-overview/oracles#zest-protocol-uses-pyth
+    oraclesBreakdown: [
+      {
+        name: "Pyth",
+        type: "Primary",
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @bats4 please use oraclesBreakdown as well to also include the proof as part of the listing, it can be added like this:
    oraclesBreakdown: [
      {
        name: "Pyth",
        type: "Primary",
        proof: ["https://docs.zestprotocol.com/start/borrow/zest-protocol-borrow-overview/oracles#zest-protocol-uses-pyth"]
      }
    ],
And should be added as a new line under the existing oralces entry

bats4:
added `oraclesBreakdown`

is this how i should do PR on other protocols for oracles as well?
even those with only one oracle in use?

thank you

realdealshaman:
Hey, @bats4 yes please, eventually this will help us add more info, adding it to the "oracles:[]" portion is only required when it is the primary or used in an aggregator. but for other types you can add them in oraclesBreakdown first. Here are some types: 
  oraclesBreakdown?: Array<{
    name: string,
    type: 

PR #9563, opened 2025-03-27
Update dTRINITY parent category and dUSD coingecko id

0xstupig:
We realized that our protocol didn’t have a category on DeFiLlama, so I updated it to the correct one for better exposure. Additionally, I updated the CoinGecko ID for our dUSD


realdealshaman:
Hey @0xstupig, thanks for the PR. The category should be added at the product level, not to the parent protocol, see other parent listings for reference. Also, we only include the coingecko id to the parent protocol listing if one exists, and this field should reference the governance token

PR #9567, opened 2025-03-28
Oracles: Add Colend to Switchboard supported protocols

mgild:
## Add Colend to Switchboard supported protocols 

Colend uses Switchboard oracles in conjunction with Pyth for robustness of pricing sources

Oracle documentation referencing how different price sources are used can be found here: 
https://colend.gitbook.io/home/resources/security-and-risks/oracles#motivations-for-using-a-pull-oracle-over-a-push-oracle


realdealshaman:
Added Switchboard as a Fallback oracle

PR #9570, opened 2025-03-28
Add Stork as oracle for EVAA

BrianGarchitorena:
Hey! Another request for you to add Stork as an oracle. It was tough to tell primary from secondary, but because Stork supports the highest number of feeds, we decided we'd submit it as primary.

Thanks!

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -34387,7 +34387,19 @@ const data3: Protocol[] = [
     module: "evaa/index.js",
     twitter: "evaaprotocol",
     forkedFrom: [],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hi @BrianGarchitorena, thanks for the PR! Based on the docs, EVAA computes a median price from all oracle providers, so we can’t select a single ‘primary.’ Instead, we should list all providers under the oracles: [] key, and in oraclesBreakdown, assign type: 'Aggregator' for each one

BrianGarchitorena:
Gotcha! Didn't realize 'Aggregator' was an option- will update now.

PR #9585, opened 2025-03-31
Update defi tuna oracle

bats4:
Hi 
made a PR to update oracle for DefiTuna


realdealshaman:
Hey @bats4 thanks for the PR! I have a question regarding the wording used in their docs. When they say Pyth is partially used, is there a way to see or somewhere to find how Pyth is being used and if there are other oracles? Mostly becaus the use of word "partially" raises question as to how their oracle system works

bats4:
hey @realdealshaman 
i would need to go back to you on that. they say they use Orca Pools price and uses Pyth to double-check.
that's only from a discord conversation though and dont have Docs about it

bats4:
hey @realdealshaman 
just noticed that DeFiTuna updated their docs, mentioning

> DeFiTuna relies on Pyth as its primary oracle

link is same, so i don't think i need to revise anything

although just remembered from a previous PR, that you asked me to use oraclesBreakdown as well even with just 1 oracle

do you want me to revise using that one?

PR #9598, opened 2025-04-01
hide unifi navbar

JasminMusic:
Hide unifi navbar


realdealshaman:
Hey @JasminMusic thanks for the PR but why hide this project?

JasminMusic:
@realdealshaman Unifi vault is already included under the Puffer Vault tab, so there's no need to display it as a separate tab.

image

Basically, we would like this tab not to be shown when we visit the https://defillama.com/protocol/puffer-finance

realdealshaman:
Hey, we'd prefer to keep it for better visibility of each product, we are already not counting it in the parent listing

PR #9599, opened 2025-04-01
feat: add the integration for sandglass price

metalgear-dev:


metalgear-dev:
How to check if the current code actually returns the correct list of the `Write` objects?

metalgear-dev:
I tried to test with the command like `npx ts-node ./src/adapters/yield/sandglass/index.ts`, but the following error occurs:

image

waynebruce0x:
I made these changes just to make things tidier https://github.com/DefiLlama/defillama-server/pull/9604/commits/e4df9c52764fda11d43d0317a387062424fcea99

PR #9624, opened 2025-04-04
feat: support Noble Dollar

johnletey:
This PR adds support for the Noble Dollar (USDN). As the collateral of the token is solely M by M^0, we choose to use their Coingecko ID until we're directly supported there. This will be paired with an adapter PR to update the TVL computation of the Noble blockchain.

### DIFF
--- a/coins/src/adapters/tokenMapping.json
+++ b/coins/src/adapters/tokenMapping.json
@@ -6160,6 +6160,11 @@
       "to": "coingecko#ondo-us-dollar-yield",
       "decimals": 18,
       "symbol": "ausdy"
+    },
+    "uusdn": {
+      "to": "coingecko#m-2",
+      "decimals": 6,
+      "symbol": "uusdn"
     }
   },
   "neox": {


coins/src/adapters/tokenMapping.json
```diff
@@ -6160,6 +6160,11 @@
       "to": "coingecko#ondo-us-dollar-yield",
       "decimals": 18,
       "symbol": "ausdy"
+    },
+    "uusdn": {
+      "to": "coingecko#m-2",
```
waynebruce0x, line 6165: This API ID wont return price data because its `priced by coingecko`

coins/src/adapters/tokenMapping.json
```diff
@@ -6160,6 +6160,11 @@
       "to": "coingecko#ondo-us-dollar-yield",
       "decimals": 18,
       "symbol": "ausdy"
+    },
+    "uusdn": {
```
waynebruce0x, line 6164: this should be a defillama chain slug

johnletey: The other tokens above are also from Noble and don't have a slug.

coins/src/adapters/tokenMapping.json
johnletey: Hmmm, the other tokens from Noble also have Coingecko?

waynebruce0x: we can merge it but i dont think this will add any price support

PR #9633, opened 2025-04-06
disable mind network

joshuamind:


realdealshaman:
Hey @joshuamind thanks for the PR but we don't delist sunsetted projects, we prefer to keep them for historical reference of TVL

PR #9673, opened 2025-04-10
fix: origin-defi to origin-protocol

ccerv1:
The parent protocol on the website is currently displayed as "https://defillama.com/protocol/origin-protocol" however the data returned for its child protocols map to "origin-defi" (which appears to be deprecated)


realdealshaman:
Hey @ccerv1 thanks for the PR! Unfortunately the id is not something we can change, it is used to link the info in our DB so changing the name will introduce unwanted changes

PR #9682, opened 2025-04-13
Update Mind Network chain info in the metadata.

joshuamind:


realdealshaman:
Hey @joshuamind thanks for the PR, if Mind Network is on Ethereum and Scroll according to the tvl adapter, then we need to keep those chains in the listing

PR #9683, opened 2025-04-13
tag deprecated protocols in /lite/protocols

mintdart:


defi/src/storeGetProtocols.ts
```diff
@@ -175,6 +175,7 @@ export async function storeGetProtocols({
       mcap: protocol.mcap,
       gecko_id: protocol.gecko_id,
       parent: protocol.parentProtocol,
+      ...(protocol.deprecated || protocol.name.includes("eprecated") ? {deprecated: true} : {})
```
0xngmi, line 178: why check protocol name?

mintdart: some protocols like rage trade v1 have deprecated in their name https://api.llama.fi/lite/protocols2

PR #9690, opened 2025-04-14
Update Datamine Network listing

hodlforjesus:
Updating Datamine Network listing to:

- Exclude address & symbol (we have 4 tokens in the ecosystem)
- Updated audits information (we had 2 audits by SlowMist). Proof: https://github.com/Datamine-Crypto/white-paper/tree/master/audits
- Added Ethereum to chains list (as our ecosystem spans L1 and Arbitrum L2)


realdealshaman:
Hi @hodlforjesus thanks for the PR! Please revert changes to audits and audit_notes, we should only include audits if the protocol itself has been audited

hodlforjesus:
@realdealshaman removed the audits, thanks!

PR #9691, opened 2025-04-15
update Desyn safe oracles

santinoYoung:


santinoYoung:
@realdealshaman Hi,team,I've had this submission for a long time with no response

realdealshaman:
Please reopen a fresh pr with info on how the oracle is being used and how it is keeping the users token safe from being stolen

PR #9699, opened 2025-04-16
b14g: PT-dualBTC support on CORE

nguyenquangthang1997:


coins/src/adapters/tokenMapping.json
```diff
@@ -6377,6 +6377,11 @@
       "decimals": 8,
       "symbol": "oBTC"
     },
+    "0xd60d60b168360dbd3496ec0ed0afe4a7498de3ce": {
+      "to": "coingecko#bitcoin",
+      "decimals": 8,
+      "symbol": "PT-dualBTC"
```
waynebruce0x, line 6383: This trades at below the value of BTC because its just principle right?

waynebruce0x: this token mapping will peg PT-dualBTC to BTC which we cant do

PR #9709, opened 2025-04-16
Add Stork as oracle for Helix

BrianGarchitorena:
Hi again,

Submitting a PR for Stork as the primary oracle for Helix. Looks like for mark prices it's split between Stork and Pyth, but for unique use-cases and index perpetuals, Helix uses solely Stork.

Thanks!

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -54695,12 +54695,12 @@ const data3: Protocol[] = [
     module: "injective-orderbook/index.js",
     twitter: "HelixApp_",
     forkedFrom: [],
-    oracles: ["Stork","Pyth"],
+    oracles: ["Pyth"],
     oraclesBreakdown: [
       {
         name: "Stork",
-        type: "Primary",
-        proof: ["https://docs.helixapp.com/trading/perpetuals"]
+        type: "Secondary",
+        proof: ["https://github.com/DefiLlama/defillama-server/pull/9709"]
       },
       {
         name: "Pyth",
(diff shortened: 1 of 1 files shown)


BrianGarchitorena:
Hey RealDealShaman. Wanted to follow up on this one to see if you guys needed any additional info.

realdealshaman:
Hey @BrianGarchitorena thanks for the PR! Can you get the Helix team to comment on this PR?

BrianGarchitorena:
@realdealshaman of course! And yeah no problem, we reached out to Helix and they said they'll have someone from the team take a look :)

ckhbtc:
Hey there - sorry for the delay, CK from Helix here. Helix utilizes a number of decentralised oracle providers including Pyth and Stork !

BrianGarchitorena:
Thanks @ckhbtc ! Just edited. Wasn't sure before, but I saw there can be multiple primary oracles so I made Stork and Pyth both primary

realdealshaman:
Hey @ckhbtc, thanks for the info! I was under the impression that Band was the primary oracle. When Pyth and Stork are also involved, are they being aggregated (e.g. averaged together), or are they used independently for specific assets? And just to clarify — is Band still the main source for most pricing, or has that shifted?

ckhbtc:
@realdealshaman No trading pairs on Injective/Helix currently use Band.

realdealshaman:
@ckhbtc got it so Band is not used, and Pyth and Stork are being used within an aggregator that uses the average between the two oracles? Is this correct?

ckhbtc:
@realdealshaman Each trading pair uses one oracle, either Pyth or Stork. Not an average of the two prices.

realdealshaman:
@ckhbtc  That makes sense, thank you! For us to determine which is the primary oracle, we need to know which one of those two is pricing the assets that make up the majority of TVL

ckhbtc:
@realdealshaman The primary would be Pyth, secondary would be Stork.

PR #9723, opened 2025-04-19
Update of echelon market & moveposition

bats4:
Hi updated oracles for Echelon Market and Moveposition

Echelon Market
- They are also in Movement Chain, so added the chain + Oracle

Moveposition
- Added oracle


bats4:
Added Monad and Movement on chainsbyoracle - Pyth

PR #9808, opened 2025-05-08
update cian oracles

NatureLab:
Kindly asking to update oracles for Cian yield layer.

Oracle Provider(s): Chainlink

Documentation/Proof:
https://docs.cian.app/getting-started-with-cian/yield-layer/introduction/chainlink-oracles

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -36416,33 +36416,6 @@ const data3: Protocol[] = [
     audit_links: ["https://dedaub.com/audits/illuminex/illuminex-jan-19-2024/"],
     listedAt: 1710199785,
   },
- /* {
-    id: "4276",
-    name: "CIAN Vaults", // previous CIAN LSTs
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey, @NatureLab thanks for the PR, but we prefer to leave the old listing in the data.ts file

Also can you tell us a bit about how Chainlink is used in name: "CIAN Automation",

NatureLab:
> Hey, @NatureLab thanks for the PR, but we prefer to leave the old listing in the data.ts file
> 
> Also can you tell us a bit about how Chainlink is used in name: "CIAN Automation",

Cian Yield layer is using Chainlink BTC and ETH feeds as security check before accepting deposits. We have made an edit to our documentation. Cian Automation however does not use any Oracle so can be removed.

realdealshaman:
@NatureLab thanks for the PR! I've added the changes manually to avoid merge conflicts. Thanks again

PR #9828, opened 2025-05-14
Add SundaeSwap V3 Popular Pools

whankinsiv:
This adapter is untested and will require modification from a DefiLlama team member. The current version of the adapter includes the correct query to collect the top 50 pools on SundaeSwap V3 and filters this list to only include pools with ADA being the other asset. Adding this adapter will allow DefiLlama to start tracking many currently untracked Cardano tokens including WMTX and MIDAS.


waynebruce0x:
Does this look right to you? https://github.com/DefiLlama/defillama-server/pull/9835

whankinsiv:
> Does this look right to you? #9835

Yes this looks great! Thank you.

PR #9830, opened 2025-05-14
euler vault pricing

darvinrio:


waynebruce0x:
will merge to debug

waynebruce0x:
apologies for the AWS dependency blocking testing. The adapter seems to work perfectly when I run it. The asset prices should now all be supported. I'll retrigger CI on the adapters test and we can check if the results are expected

darvinrio:
sounds amazing

PR #9842, opened 2025-05-15
feat: add eOracle to Euler's oracle list

totomanov:


realdealshaman:
Hey @totomanov thanks for the PR! How much of Euler V2 total TVL is eOracle securing? We can calculate this from the assets they are pricing with eoracle feeds and the amount of TVL they make up

totomanov:
All curators/governors on Euler can choose their own oracle for their vaults. eOracle is one of several providers supported - they are all visible here https://oracles.euler.finance/1/

context: https://github.com/DefiLlama/defillama-server/pull/8231

realdealshaman:
Hey, thanks for the pr! like mentioned in the discord channel, we've removed the oracles from Euler since they don't set an oracle themselves, the oracle should be added to the risk curator that chooses it

PR #9853, opened 2025-05-15
added eoracle

leemount96:


realdealshaman:
Hey, @leemount96 thanks for the PR! How much of Euler V2 total TVL is eOracle securing? We can calculate this from the assets they are pricing with eoracle feeds and the amount of TVL they make up

leemount96:
At the moment they are supporting a few oracles on Bera, but have only recently been onboarded and expect new markets/TVL to be added. TVL is ~$5-10mm at the moment but again this will grow

PR #9870, opened 2025-05-19
keep oracle even if not active, with TVL 0 so UI can display, split aggregator correctly for historical data

ReynardoEW:


defi/src/protocols/data.ts
```diff
@@ -8455,6 +8455,7 @@ The eWIT token is a custodial, wrapped version of the Witnet coin managed by the
     chains: ["Tron"],
     module: "justlend.js",
     twitter: "DeFi_JUST",
+    oracles: ["WINkLink", "Chainlink"],
```
0xngmi, line 8458: this has been superseded by oraclesBreakdown, oracles should no longer be used here

ReynardoEW: currently http://github.com/DefiLlama/defillama-server/blob/master/defi/src/storeGetProtocols.ts#L67 relies on oracles directly, so we should map from oraclesBreakdown to oracles on storeGetProtocols instead? https://github.com/DefiLlama/defillama-server/blob/master/defi/src/protocols/data3.ts#L45064 also using oracles and oraclesBreakdown at the same time, right now the merged code of getOracles splitting the TVL based on the oracles breakdown but the protocol wont show up (the tvl do show up though) https://defillama.com/oracles/Coingecko/Aptos

mintdart (approved):

PR #9874, opened 2025-05-20
Update Solera chain from Plume (Deprecated) to Plume Mainnet

leoluxembourg:
Updated metadata


realdealshaman:
Hey @leoluxembourg thanks for the PR! there were a couple of errors in the new metadata, I've fixed these manually and added the new description, and x.com handle along with the new audits. For the url we need to keep the url for the actual dapp, so I have removed that change. Thanks again, will close this PR for now since the new info has been added

PR #9892, opened 2025-05-21
Add Stork as a fallback oracle for Orderly

BrianGarchitorena:
Hey @realdealshaman Adding in Stork as a fallback oracle for Orderly.

Thanks!


realdealshaman:
updated the listing to keep both oracles as Fallback

PR #9927, opened 2025-05-31
update synstation information

jooddang:
I'm the maintainer of SynStation. Because the information is outdated, I'm updating the project information of SynStation. Please review and approve.
Thanks!


realdealshaman:
Hey @jooddang, we need to keep the current category as-is unless a new TVL adapter is created. We also can’t update the chain until the adapter reflects the correct info. Based on your request, it looks like this is a completely different version of the project, so we’ll likely need to a new listing for it

PR #9949, opened 2025-06-08
check for l2 and add to parent chain app fees

ReynardoEW:
DefiLlama/defillama-app#1922


0xngmi:
can this be closed now? if so pls close it

PR #9974, opened 2025-06-11
Add DIA Oracles to Peapods Finance

Josh-DIA:
Hello DefiLlama team,

Requesting to add Peapods Finance to DIA Oracles TVS.

Oracles: DIA
Implementation Details: Enabling external lending and borrowing protocol integrations for the Peapods Finance native PEAS token (on all markets). Includes additional price feeds for SERV, NPC, XMW.
Documentation: Peapods request on forum (https://forum.diadata.org/t/cdr-091-peapods-finance-asset-price/1011/5), partnership details (https://www.diadata.org/blog/post/partnership-with-peapods-finance/)
Failure Scenario: Stale price data can lead to vaults being liquidated (as these are leveraged, more susceptible to price variation), which can result in the TVL of each vault being under risk.

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -30423,6 +30423,7 @@ const data3_1: Protocol[] = [
     twitter: "PeapodsFinance",
     audit_links: ["https://sourcehat.com/audits/PeapodsFinance/"],
     listedAt: 1706184132,
+    oracles : ["DIA"],
   },
   {
     id: "4064",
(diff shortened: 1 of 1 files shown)


Josh-DIA:
Hey @realdealshaman, yes that is correct they are using our oracle to secure their LVF TVL on their protocol. Thanks!

realdealshaman:
Got it! Thanks for clarifying! For future entries, we need to use oraclesBreakdown instead of oracles, which we are winding down

Josh-DIA:
Understood, thanks!

PR #10047, opened 2025-06-24
Update oracles for Euler on Unichain & Swellchain.ts

Hundert1000:
Hey Llamas,

Kindly asking to add RedStone as the primary oracle for Euler on Unichain & Swellchain. RedStone is the only integrated and used oracle by Euler on those chains.

Documenation/Proof:

Unichain -> https://oracles.euler.finance/130/

Swellchain -> https://oracles.euler.finance/1923/

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -52585,6 +52585,20 @@ const data3_2: Protocol[] = [
     cmcId: null,
     category: "Lending",
     chains: ["Ethereum"],
+    oraclesBreakdown: [
+      {
+        name: "RedStone",
+        type: "Primary",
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @Hundert1000 I believe these oracles are not being set by Euler but instead they are set by the risk curators

Hundert1000:
Hey @realdealshaman n — again you are right, the oracles in these markets are indeed set by the risk curators.

In this case, RedStone was selected as the Oracle, and these markets currently entirely and exclusively rely on RedStone price feeds. That means the associated TVL is secured by RedStone, and as such, we believe it should be listed under RedStone TVS.

Let me know if there’s anything we might be missing here or if any additional clarification would be helpful!

PR #10070, opened 2025-06-28
Update data.ts to update Paintswap

DoughBender:
Updated Paintswap url, description, category + added Sonic network.


realdealshaman:
Hey @DoughBender, thanks for the PR! A few things can’t be changed:
	•	The chain: "Fantom" field reflects the original deployment chain and shouldn’t be updated.
	•	We can’t add multiple categories or contract addresses.
	•	Since the TVL code hasn’t changed, the category must remain the same — to use nft marketplace, the adapter needs logic for NFT-specific data.
	•	We also can’t include Sonic in chains: [] until it’s supported in the TVL adapter. We have checks to prevent adding unsupported chains.

Please adjust the PR accordingly — thanks!

DoughBender:
Greetings @realdealshaman 

Thanks for the feedback, and appreciate the clarification.

Just to provide more context:

Fantom has rebranded to Sonic, and Paintswap migrated its core infrastructure along with the BRUSH token to Sonic. The BRUSH token currently exists on both Fantom and Sonic, but Sonic is expected to fully phase out Fantom over time. Paintswap on Sonic is live and actively supported, with distinct contracts and growing TVL.

Our goal was to ensure Sonic gets properly reflected in the metadata to help users discover where Paintswap is actively operating today. That said, we understand chain should reflect the original deployment, and that chains requires adapter support. 

How can we get about providing Sonic support to our BRUSH token? Should we start a separate token listing for the BRUSH in Sonic?

Will appreciate any lead. Thank you

DoughBender:
Hello @realdealshaman . 

Apologies of the lapse. We do have this endpoint: https://api.paintswap.finance/v2/stats/
Can we use this with the adapter and make a new listing for "Paintswap Sonic"?

realdealshaman:
@DoughBender we can just close it, once the new adapter is ready, I'll do the new listing, update the names and add them both under a PaintSwap parent listing

DoughBender:
Okay. That sounds like a plan. Will close this for now and will do a PR on /adapters. will keep this PR as reference. 

Thank you very much for your assistance.

PR #10090, opened 2025-07-02
Mark-dead-urls

FrozenEcho44:
Set `deadUrl: true` for 475 projects with parked or unreachable domains.

Detection methodology:

129 URLs identified by parking page patterns
27 URLs identified by content analysis
303 URLs identified by network errors (220 DNS resolution failures, 72 certificate errors, 11 SSL protocol errors)

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -10749,15 +10749,16 @@ const data3_1: Protocol[] = [
   },
   {
     id: "3179",
-    name: "GoGoPool",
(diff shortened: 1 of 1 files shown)


defi/src/protocols/data3.ts
```diff
@@ -10749,16 +10749,15 @@ const data3_1: Protocol[] = [
   },
   {
     id: "3179",
-    name: "Hypha",
```
0xngmi, line 10752: why this change?

PR #10128, opened 2025-07-09
add satUSD v2 to BOB

mastilver:


realdealshaman:
Hey @mastilver thanks for the PR, but adding this entry isn't needed since we already track the price of that address:

https://coins.llama.fi/prices/current/bob:0xecf21b335B41f9d5A89f6186A99c19a3c467871f

PR #10134, opened 2025-07-10
Update glif project info

Schwartz10:


realdealshaman:
Hey @Schwartz10 thanks for the PR! To add a chain to the listing we need to add it to the adapter first: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/glif/index.js

Do you have a link to the audits, it could be in a github file or a link to the audits page in the docs? We don't actually use the audit_note field in the listing so we'll need to revert those changes please

PR #10174, opened 2025-07-14
correct hOHM token mapping

frontier159:


frontier159:
> Right now we havent got code for finding prices from kodiak pools

Ah ok - it's a fork of uniswap v3. Do you have that code available and can you link me to it? 

I can ask the Kodiak team to implement - it might really help save some of your time!

PR #10196, opened 2025-07-16
Add oracle for xstocks and dead url true for gmd protocol

KPHEMRAJ:
Source : https://x.com/xStocksFi/status/1942192950882259303 and https://x.com/xStocksFi/status/1939761143922532548?t=vF7ER9vvW-2Rrm6cNbcU6A&s=19

### DIFF
--- a/defi/src/protocols/data2.ts
+++ b/defi/src/protocols/data2.ts
@@ -21461,6 +21461,7 @@ const data2: Protocol[] = [
     audit_links: ["https://solidity.finance/audits/GMDAOVault"],
     listedAt: 1669732077,
     governanceID: ["snapshot:gmdgovernance.eth"],
+    deadUrl: true
   },
   {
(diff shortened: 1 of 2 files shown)


realdealshaman:
Hey @KPHEMRAJ thanks for the PR! For new oracle entries we need to use oraclesBreakdown. We have more info on it here:
https://docs.llama.fi/list-your-project/oracles-tvs#explanation-of-oraclesbreakdown

KPHEMRAJ:
> p.s. we need to know how each oracle is working within the project so that we can set they to the correct type: Primary, Secondary, Fallback, Aggregator etc

Hey @realdealshaman thanks for the info. I am removing oracle part bcz of insufficient data, can we go ahead as gmd protocol url is dead

realdealshaman:
thanks for the PR

PR #10212, opened 2025-07-17
Fix StakeDAO Protocol Listing Information

warrenception:


realdealshaman:
Hey @warrenception thanks for the PR! To add Fraxtal and Base to the chains mapping, they need to be added to the projects adapter first. Here: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/stakedao/index.js

PR #10220, opened 2025-07-21
validate api script (#10217)

ReynardoEW:
uses api-docs schemas, some error caused by wrong schemas definition so need to make changes there

.env setting
BASE_API_URL=https://api.llama.fi
BETA_API_URL=
DISCORD_WEBHOOK_URL=
PRO_API_KEY=

### DIFF
--- a/defi/package.json
+++ b/defi/package.json
@@ -57,7 +57,11 @@
     "twitter-update": "npx ts-node --logError --transpile-only src/twitter/updateTwitterData.ts",
     "local-test-por": "npx ts-node --logError --transpile-only proof-of-reserves/cli/test.ts",
     "run-check-por": "npx ts-node --logError --transpile-only proof-of-reserves/cli/check.ts",
-    "ui-tool": "npm run update-submodules; cd ui-tool; npm run start-server"
+    "ui-tool": "npm run update-submodules; cd ui-tool; npm run start-server",
+    "validate-api": "npx ts-node --logError --transpile-only src/utils/validateAPI.ts",
+    "validate-api-pro": "npx ts-node --logError --transpile-only src/utils/validateAPI.ts --api-type pro",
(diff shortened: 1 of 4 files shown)


treeoflife2 (approved):

treeoflife2:
@ReynardoEW are we sending the json file in the discord? as discord message will be truncated when it's very big? everything else looks good.

ReynardoEW:
> @ReynardoEW are we sending the json file in the discord? as discord message will be truncated when it's very big? everything else looks good.

it send summaries not full json file something like this

```api validation error

pro api:
• total endpoints: 82
• pass: 77
• fail: 5
• success rate: 93.9%
• avg resp time: 1401ms

failed endpoints (5 of 5):
1. https://pro-api.llama.fi/api/emission/hyperliquid
2. https://bridges.llama.fi/bridgevolume/Ethereum?id=5
3. https://bridges.llama.fi/bridgedaystats/1667304000/Ethereum
4. https://bridges.llama.fi/transactions/1?starttimestamp=1667260800&endtimestamp=1667347200&sourcechain=Polygon&address=ethereum%3A0x69b4B4390Bd1f0aE84E090Fe8af7AbAd2d95Cc8E&limit=200
5. https://pro-api.llama.fi/coins/block/sample/sample

2025-07-21T06:19:06.323Z```

treeoflife2:
@ReynardoEW it will only send the alert if there is failure right? imo json file would be better for debugging, but we can add it later if needed for jenkin jobs, or we can add link of file, as i believe filename will always same we can later use static link, but fine for now.

treeoflife2:
@ReynardoEW can we also automate the generation of this llms.txt file, it uses only pro-apis request/response information from the defillama-pro-openapi.json, everything else will stay the same
https://raw.githubusercontent.com/DefiLlama/api-docs/refs/heads/main/defillama-llms.txt

PR #10250, opened 2025-07-23
digift

waynebruce0x:


coins/src/adapters/rwa/digift.ts
```diff
@@ -41,5 +41,6 @@ export async function digift(timestamp: number = 0) {
   const writes: Write[] = [];
   await getTokenPrices("ethereum", timestamp ,writes);
   await getTokenPrices("arbitrum", timestamp, writes);
+  await getTokenPrices("plume", timestamp, writes);
```
ckp-0217, line 44: Could you help change "plume" to "plume_mainnet"? We have already switched the chain in DefiLlama-Adapters.

PR #10279, opened 2025-07-27
Update `Blueberry` Description

Gornutz:
Update to reflect the relaunch of blueberry now on Hyperliquid L1.


realdealshaman:
Hey @Gornutz you need to create a new adapter for the new version and we will need to do a new listing for it that we can name blueberry v2

PR #10315, opened 2025-07-31
feat: add `cBABY` to token mapping

gjj:
As per TG discussion and https://github.com/DefiLlama/defillama-server/pull/10305, adding in cBABY mapping to the CoinGecko page.


gjj:
resolved! @waynebruce0x for re-review

PR #10325, opened 2025-08-01
SurgeDefi oracle update

surgedefi:


realdealshaman:
@surgedefi thanks for the PR but we skip adding the oracle to Dexs since they don't rely on pricefeeds for tokens the way a lending or perps project would

PR #10385, opened 2025-08-12
Updated description for Rezerve and set DIA as our primary oracle

drecronye:
Docs updated over https://rezerve.gitbook.io/protocol/security/oracles


realdealshaman:
Hey @drecronye thanks for the PR! but we can simply update the name of the primary oracle, we need to set an end date for when eoracle stopped being the primary and a start date to a new entry below it for DIA

PR #10393, opened 2025-08-13
add sosovalue token details

KPHEMRAJ:


realdealshaman:
Hi @KPHEMRAJ thanks for the PR! I've added the token address and ticker manually to included in all the child listings and I've added the coingecko id and cmc id to the parent listing. Thanks again

PR #10399, opened 2025-08-14
add Pyth to protocols

mariobern:
gm llama sirs

adding 2 new protocols

Hylo
Defi Dollar

Proofs attached in the commit

Thank you


realdealshaman:
Hey @mariobern, thanks for the PR! From the docs, most tokens locked in Defi Dollar are priced by Chainlink. We’d need to list Pyth as secondary unless we can show or provide an argument/case that a mispricing in any Pyth-priced token could drain over 50% of the TVL
image
 From this list only LQTY is using Pyth

PR #10415, opened 2025-08-15
update stake.link

BkChoy:
- removed stake.link index as it's no longer supported
- removed parent protocol as there's no need for it anymore with the removal of stake.link index
- updated and renamed stake.link liquid to be the sole page for stake.link


realdealshaman:
Hi @BkChoy thanks for the PR but we prefer to keep deprecated product listings for the historical data. We can add a deprecated flag to the project listing though so that it is noted

PR #10424, opened 2025-08-18
add solidex-finance-v2 and solidex-finance-v3

solidexdev:


realdealshaman:
Hey, we missed this PR but the listings have been added. Thanks for the info and the help

PR #10433, opened 2025-08-18
dedup keys of chainTvlsToAdd in storeTvl

kws1207:
While investigating below issue, I noticed that in `storeTvl` function, `chainTvlsToAdd` uses a `string[]` as its value type. This can lead to incorrect behavior if duplicate keys are inserted for any reason.
- https://github.com/DefiLlama/DefiLlama-Adapters/issues/15789

I updated it to use a `Set` instead of `string[]` to ensure uniqueness.


aminakausar75-crypto:
Now

aminakausar75-crypto (approved):
Approve

aminakausar75-crypto (changes requested):
Request changes

aminakausar75-crypto (approved):
Merging these changes

PR #10452, opened 2025-08-20
Update oracle for Avalon USDa.ts

Hundert1000:
Hey Llamas,
Kindly asking to update oracle for Avalon USDa across all chains.

Implementation Details: Avalon uses Redstone as its primary oracle for the BTC-USD market securing our BTC-backed stablecoin, USDa, across all supported networks.

Documenation/Proof: https://docs.avalonfinance.xyz/avalon-products/cedefi-cdp-usda/risk-management#oracle-solutions

### DIFF
--- a/defi/src/protocols/data3.ts
+++ b/defi/src/protocols/data3.ts
@@ -58647,7 +58647,7 @@ const data3_2: Protocol[] = [
     cmcId: null,
     category: "CDP",
     chains: ["Ethereum", "Binance", "Mantle"],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hi @Hundert1000 thanks for the PR! I've updated the listing according to the proof you've provided. We should not delete old data anymore for oracle listing, we can simply add an endDate for the existing oracle and a startDate for the new primary. Like this:
image

PR #10453, opened 2025-08-20
Name change on canary protocol metadata

hugofqueiros:
We're requesting a name changing as the protocol is legacy at this point.

thank you for your time!


realdealshaman:
Hi @hugofqueiros thanks for the PR! Is there a public announcement that can be used as a reference of the name change?

PR #10496, opened 2025-08-25
update search list script

mintdart:


mintdart:
- updated to check if `route` exists instead of `id`
- moved code related to getting items to delete from db to a seperate function

defi/src/updateSearch.ts
```diff
@@ -249,10 +268,111 @@ const getProtocolSubSections = ({
...
+async function getAllCurrentSearchResults() {
+  const res: { total: number; results: Array<SearchResult> } = await fetch(
+    "https://search.defillama.com/indexes/pages/documents",
+    {
+      headers: {
+        Authorization: `Bearer ${process.env.SEARCH_MASTER_KEY}`,
+      },
+    }
+  ).then((res) => res.json());
+  const finalRes: { total: number; results: Array<SearchResult> } = await fetch(
```
0xngmi, line 280: i dont think this will work, what you need to do is fetch the new page in a loop till the number of items returned is smaller than pagesize

this is basic pagination

defi/src/updateSearch.ts
```diff
@@ -249,10 +268,111 @@ const getProtocolSubSections = ({
...
+        return true;
+      }
+
+      if (item.type === "Stablecoin") {
+        return true;
+      }
+
+      // it is okay to delete as this relies on frontend pages, can't really be sure if it is still valid
+      if (item.type === "Metric") {
+        return true;
```
0xngmi, line 341: isnt this gonna create a similar issue as we had before, where if one query succeeds but the other doesnt it will lead to missing results?

0xngmi (changes requested):

PR #10507, opened 2025-08-27
Update Yala Oracle

bats4:
HI
made a PR on Yala based on docs

Thank you


realdealshaman:
Hey @bats4 thanks for the PR! I see that we should update the oracle section for Yala but do you have any info stating that a median or mean is being used from the Pyth and Switchboard feeds? I can see that both are actively used, but neither is confirmed as the primary source

bats4:
only got this answer
image

PR #10509, opened 2025-08-27
Updated Echelon oracles

femedmad:
Hi DefiLlama team! Echelon is now being secured by Chainlink oracles! You can find the official announcement [here](https://x.com/EchelonMarket/status/1960734061938532857) and the proof [here](https://docs.echelon.market/echelon/core/oracles).


realdealshaman:
@0xbe1 please explain how we should read the frontend because most markets appear with both chainlink and pyth but some have pyth on it's own and and in other cases we see the pyth icon with the switchboard icon, also chainlink does not appear in any of the movement pools:
image

PR #10546, opened 2025-09-02
Jarryd/add figure heloc token

jarrydallison:
Removes the old pm.sale.pool token, and adds FIGR_HELOC token (https://www.coingecko.com/en/coins/figure-heloc). Also, enables YLDS instead of converting it to USD-COIN (https://www.coingecko.com/en/coins/ylds)


waynebruce0x:
NFTs must be added here instead https://github.com/DefiLlama/defillama-server/blob/master/coins/src/adapters/zeroDecimalTokenMapping.json

jarrydallison:
Sorry @waynebruce0x I didn't realize this never merged. Updated.

PR #10623, opened 2025-09-16
Update data2.ts

Eteriaio:


realdealshaman:
Hey, we need to add new listings to the most recent data.ts file, currently that is data4.ts
I've added the listing, it will be live shortly

PR #10632, opened 2025-09-17
Update Oracles for Termmax.ts

isaacRedstone:
Hey Llamas,
Kindly asking to update oracle for Termmax.

Implementation Details: "Termmax integrated RedStone price feeds as our primary source of data for assets, we leverage RedStone oracles for main pools including USDC, wstETH & WETH."

Documenation/Proof: https://docs.ts.finance/technical-details/oracles


realdealshaman:
Hey @isaacRedstone thanks for the PR! As I look through the docs I see that RedStone is mentioned as the primary oracle, however when we check the mapping below, we see that Chainlink is mapped as the oracle for WSTETH and for the pendle tokens

PR #10651, opened 2025-09-19
Update MorpheusAI info

antonbosss:


realdealshaman:
hey @antonbosss thanks for the PR! Please add only one address for the token, our scripts only support the use of one

Also to add new chains to the chain mapping, the chains need to be added to the tvl adapter first. We can get the PR merged once these additional changes take place. Thanks

PR #10672, opened 2025-09-23
Update data for Zest Protocol: added DIA to oracles

y0s0n:
Updated the oracle breakdown to reflect that we are using DIA oracles for some assets.


realdealshaman:
Hey @y0s0n thanks for the PR! Can you tell us a bit about what is being shown and how DIA is used within Zest?

PR #10702, opened 2025-09-29
Delist: O.LAB

nikli2009:
Hi Defillama team, 
Please help us delist our entry. 🙏

Thanks in advance


realdealshaman:
@nikli2009 I've flagged it as deprecated, it will show with the next hourly update

PR #10742, opened 2025-10-02
Etherflip category updated to Gaming instead of Prediction Market

Deantoshi:
Updated category identified in DefiLlama-Adapters/issues. https://github.com/DefiLlama/DefiLlama-Adapters/issues/16135


Deantoshi:
Updated Arrow Market description from issue identified here: https://github.com/DefiLlama/DefiLlama-Adapters/issues/15888

realdealshaman:
@Deantoshi thanks for the PR! However, we add Prediction Market to any protocol that has to do with gambling, even if it is only dice

PR #10774, opened 2025-10-07
read write functions

waynebruce0x:


defi/src/utils/shared/distressedCoins.ts
```diff
@@ -1,32 +1,91 @@
...
+    index: "distressed-assets-store*",
+    body: {
+      query: {
+        match: { _id },
+      },
+    },
+  });
 
-  return false;
+  if (isLocalClient) await client?.close();
```
g1nt0ki: can leave the connection open, can close it once when process exits, noticed this pattern in other places as well, not sure if it is a good idea to open & close connection per requests, most of our scripts run under an hour, so dont see why we need to close it per write

coins/src/adapters/bridges/index.ts
```diff
@@ -129,6 +129,7 @@ import { getCurrentUnixTimestamp } from "../../utils/date";
 import produceKafkaTopics from "../../utils/coins3/produce";
 import { chainsThatShouldNotBeLowerCased } from "../../utils/shared/constants";
 import { sendMessage } from "../../../../defi/src/utils/discord";
+import { isDistressed } from "../../utils/shared/distressedCoins";
```
g1nt0ki: I dislike the logic duplication here. 

Can you consolidates all the writes to a single function `writeToDB()`

  - this would do all the validation and filtering
  - then write to both ddb & kafka
  
this way, say I want to write something from the ui-tool or anywhere else, I know what to call, a single function that takes care of validation and storage, and in the future, we want to migrate to some other db, we update a single function

defi/src/storeTvlInterval/computeTVL.ts
```diff
@@ -215,6 +227,100 @@ async function getTokenData(readKeys: string[], timestamp: string | number): Pro
   }
 }
 
+const mcapCache: { [PK: string]: any } = {}
+
+async function getMcapData(readKeys: string[], timestamp: string | number): Promise<any> {
```
g1nt0ki: looks good, can you move this and the price function to `coinsApi.ts` in the same folder, these functions can be used elsewhere as well which have nothing to do with tvl

g1nt0ki: hmm, on second thought, if you use latest sdk version, it already has functions that return cached price & mcap info?

defi/src/utils/shared/distressedCoins.ts
```diff
@@ -0,0 +1,136 @@
...
+        },
+      },
+    },
+  });
+
+  return hits.map((hit: any) => hit._source);
+}
+
+// store distressed metadata to DDB SK0 for metadata retrieval
+export async function storeDistressedCoins(keys: string[], coinsESClient?: any) {
```
g1nt0ki: lets add a new field `isDistressed` to the price object, so the things we want to set as distressed, we set that flag true, 

so here, for a given key list, we pull current metadata objects (SK=0 on ddb), set the flag and call `writeToDB` which will take care of storing

defi/src/utils/shared/distressedCoins.ts
```diff
@@ -0,0 +1,136 @@
...
+// initial write all to temp logs index
+export async function logDistressedCoins(keys: any[]) {
+  for (const data of keys) {
+    data.reportTime = Math.floor(Date.now() / 1000);
+    await elastic.writeLog("distressed-assets", data);
+  }
+}
+
+// get list of possible distressed coins from ES logs in the last week
+export async function readDistressedLogs() {
```
g1nt0ki: looks good

defi/src/utils/shared/distressedCoins.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function sanitizeKey(key: string): string {
+  const chain = key.split(":")[0];
+  const address = key.substring(chain.length + 1);
+  const normalizedAddress = chainsThatShouldNotBeLowerCased.includes(chain) ? address : address.toLowerCase();
+  return `${chain}:${normalizedAddress}`;
+}
+
+// batch check if a list of coins are distressed
+export async function batchIsDistressed(keys: string[]) {
```
g1nt0ki: can you add the isDistressed field to the price api response, I will update the sdk to include that in getPrices response

then, for the batch call here, we check if the metadata response has `isDistressed` set to true

PR #10815, opened 2025-10-14
added gecko ids

DaniiRix:


realdealshaman:
Hey, these are already being tracked in the parent protocol, if we add them to the data.ts files our backend will count them as duplicates. I've added Superfluid manually but it is still not live

PR #10816, opened 2025-10-14
add (Strata) Tranches product

alxstrata:
Hello Team,

As we've launched the Tranches product, I'm adding here the organizational structure - the parent project and the Tranches product - along with the Tranches description.

Best, Alex


realdealshaman:
Hey @alxstrata thanks for the PR, I'll add the listing manually to avoid the merge conflicts

PR #10823, opened 2025-10-16
feat: update bitlayer name and category

haiifeng:


haiifeng:
> Hi @haiifeng thanks for the PR! Can you provide a bit of context on why the new category is more correct please

Hi @realdealshaman 

Bitlayer's current business focus is on the YBTC Family of Assets. We view the bridge as a supporting tool—much like Threshold Network's tBTC or Bedrock's uniBTC—while the core protocol functions as a decentralized BTC asset. Users can mint liquid Layer 2 BTC (Bitlayer BTC) by depositing native BTC via the bridge. To better reflect this emphasis, we propose the following updates:  
Rename the protocol from "Official Bridge" to "Bitlayer YBTC Family."  
Change the category to "Anchor BTC."  
Exclude the non-BTC portion from TVL calculations.

For more details, please refer to: https://docs.bitlayer.org/docs/YBTCFamily/bl_btc_overview

PR #10846, opened 2025-10-20
feat: add UltraYield and change Edge Capital to UltraYield by Edge Capital

trifonovtema:
Change Risk Curator: Edge Capital -> Ultrayield by Edge Capital
Add Onchain Capital Allocator: Ultrayield


realdealshaman:
hey @trifonovtema thanks for the PR but we can't change the chain field, if new chains are added to the adapter then we need to update only the chains array. Also we can't change the name of the dimensions adapter or the module. Please revert those changes.

Also why create another listing if it is only a rebrand?

trifonovtema:
> hey @trifonovtema thanks for the PR but we can't change the chain field, if new chains are added to the adapter then we need to update only the chains array. Also we can't change the name of the dimensions adapter or the module. Please revert those changes.
> 
> Also why create another listing if it is only a rebrand?

@realdealshaman  Thank you for the feedback!
I understand now about the chain field and adapter naming — I’ll revert those changes.

Just to clarify: how should I correctly represent UltraYield as an Onchain Capital Allocator and UltraYield by Edge as a Risk Curator?
Should this be a single listing with both categories under one adapter, or two separate records (one per category)?

realdealshaman:
@trifonovtema thanks for the feedback, I've added the listing manually to avoid merge conflicts. Thanks again. Any additional feedback, please comment in the telegram group. Thanks again

PR #10854, opened 2025-10-21
standardize lombard naming; rename Lombard Vault -> Lombard Vaults; u…

VaitaR:
…pdate vault category and add Base & Binance chains


realdealshaman:
hey @VaitaR thanks for the PR! We need to think of a different naming scheme, differentiating between Lombard parent protocol and Lombard as a product with a space at the end of the name is not a very good UX

VaitaR:
See it, will provide a new version

PR #10866, opened 2025-10-22
ore metadata update

iocozmo:


iocozmo:
Updated @realdealshaman, thanks for ur help

PR #10867, opened 2025-10-23
feat: add DFS Network swap v2 and v3 protocols on XLayer

Judy098:
- Add DFS Network swap v2 protocol configuration
- Add DFS Network swap v3 protocol configuration
- Both protocols deployed on XLayer network
- Inherit from DFS Network parent protocol


realdealshaman:
added these listings, the previous DFS Network listing should remain unchanged for historical data, XLayer is not included in its adapter

PR #10896, opened 2025-10-28
Lombard/names fix

VaitaR:


realdealshaman:
Hey @VaitaR thanks for the PR! We are only tracking two chains in the "Lombard" adapter:
image
Please remove any chains that are not in the tvl adapter

PR #10897, opened 2025-10-28
superior distressed coin 

waynebruce0x:


coins/src/adapters/yield/pendle/pendle.ts
```diff
@@ -322,7 +322,7 @@ async function getAllTokenInfos(chainId: number) {
   }
 
   function filterMarketData(resp: { markets: any[] }) {
-    const markets = resp.markets.filter((m: any) => (m.details?.liquidity ?? 0) > 1e5); // filtering out small markets
```
g1nt0ki: isnt 10k a good filter? 1k is too low, why this change?

waynebruce0x: I've decreased it from 100k to 10k because there are pools of around 30k TVL which morpho need prices for

g1nt0ki: ah, it is 10k, my bad

PR #10945, opened 2025-11-05
Coins decimals type 

waynebruce0x:


coins/src/scripts/coingecko.ts
```diff
@@ -285,7 +285,7 @@ async function getAndStoreCoins(coins: Coin[], rejected: Coin[]) {
               PK: normalizedPK,
               SK: 0,
               created,
-              decimals,
+              decimals: Number(decimals),
```
g1nt0ki: can you verify if this field exists, and is valid number after this transform before transforming it

coins/src/adapters/bridges/index.ts
```diff
@@ -226,6 +226,7 @@ async function _storeTokensOfBridge(bridge: Bridge, i: number) {
 
       if (i && !decimals) return;
       if (!symbol) return;
+      decimals = Number(decimals)
```
g1nt0ki: same here

PR #10969, opened 2025-11-11
Update FOOM.Cash category

foom-cash:
Update FOOM.Cash category to _Privacy_; the protocol needs to be classified under _Privacy_ instead of _Luck Games_, since its primary focus is on privacy features. Because only one category can be selected on DefiLlama, _Privacy_ is the more accurate choice. As a zero-knowledge product, FOOM.Cash is fundamentally privacy-first.

### DIFF
--- a/defi/src/protocols/data5.ts
+++ b/defi/src/protocols/data5.ts
@@ -374,7 +374,7 @@ const data5: Protocol[] = [
     audit_note: null,
     gecko_id: null,
     cmcId: null,
-    category: "Luck Games",
+    category: "Privacy",
     chains: ["Ethereum", "Base"],
     forkedFrom: [],
     module: "foom-cash/index.js",


realdealshaman:
hey, @foom-cash thanks for the PR! Although privacy is a core feature, the purpose of the protocol isn’t to provide privacy as a service, it’s to run a lottery game

foom-cash:
> hey, @foom-cash thanks for the PR! Although privacy is a core feature, the purpose of the protocol isn't to provide privacy as a service, it's to run a lottery game
 
Hi DefiLlama, @realdealshaman,

We're a serious team – our protocol is an enhanced version of _TornadoCash_, designed for significantly higher privacy and real-world utility. Technically, the ZK-proof is decomposed into individual bits. These bits can optionally serve as elements of a lottery-like game – but the lottery itself is completely **optional**. Its purpose is to create an economic model that incentivizes addresses not solely interested in privacy to actively participate in the protocol (thus driving liquidity and adoption).

I want to emphasize two key points:
* The core utility of FOOM.Cash is privacy. The lottery is not required for anonymity – it's simply an additional module.
* By decomposing ZK-proofs into bits (0/1), the linkage between addresses becomes millions of times harder than in classic TornadoCash. Even with transaction analysis, entropy (information dispersion) is much higher – a payout of $700k could theoretically originate from a deposit of just $1.

We also offer flexibility for users: participation in the lottery is optional, and those who prefer can "guarantee" a winning outcome by setting the parameter to 100%, effectively purchasing a "winning ticket."

Finally, we're backed by one of the largest private ETH treasuries:
https://defillama.com/protocol/foom?treasury=true&tvl=false.

If you have any technical questions or would like to see our whitepaper or architecture diagrams – we'll be happy to share them.

Best regards,
The Foom Team

PR #10979, opened 2025-11-12
ember

waynebruce0x:


coins/src/adapters/yield/ember.ts
```diff
@@ -0,0 +1,34 @@
...
+  const res = await axios.get(
+    "https://vaults.api.sui-prod.bluefin.io/api/v1/vaults/coins/price"
+  );
+  const writes: Write[] = [];
+
+  await runInPromisePool({
+    items: res.data,
+    concurrency: 5,
+    processor: async (m: any) => {
+      const { decimals, symbol } = await getTokenInfo(m.coinType);
```
g1nt0ki: would be nice if we cached it, or check against metadata logs pulled from ES, if it exists there, use that symbol and decimals, else pull it here

g1nt0ki (approved):

PR #10992, opened 2025-11-17
axlp

waynebruce0x:


coins/src/adapters/index.ts
```diff
@@ -174,4 +174,7 @@ export default {
   goblin: require("./markets/goblin"),
   ember: require("./yield/ember"),
   suirewards: require("./markets/suirewards"),
+  axlp: require("./liquidStaking/axlp"),
+  morphoBlue: require("./moneyMarkets/morpho"),
+  lista: require("./moneyMarkets/morpho"),
```
g1nt0ki: if morphoBlue has multiple adapters, better to do something like what we have for euler: 
image

coins/src/adapters/moneyMarkets/morpho.ts
```diff
@@ -0,0 +1,277 @@
...
+    );
+  });
+
+  return writes;
+}
+
+async function getListaVaults(chain: string) {
+  const {
+    data: { list: vaults },
+  } = await fetch(listaConfig[chain].vaultInfo).then((r) => r.json());
```
g1nt0ki: use cached Fetch so the price adapter will continue to work even if api gets changed

coins/src/adapters/moneyMarkets/morpho.ts
```diff
@@ -0,0 +1,277 @@
...
+      "https://api.lista.org/api/moolah/vault/list?page=1&pageSize=1000",
+  },
+  ethereum: {
+    vault: "0xf820fB4680712CD7263a0D3D024D5b5aEA82Fd70",
+    vaultInfo:
+      "https://api.lista.org/api/moolah/vault/list?page=1&pageSize=1000&chain=ethereum",
+  },
+};
+
+async function fetchMorphoVaultAddresses(chainId: string) {
```
g1nt0ki: can you wrap this in cached fetch as well? (you can pass this as fetcher)

coins/src/adapters/moneyMarkets/morpho.ts
```diff
@@ -0,0 +1,277 @@
...
+    processor: (vault: string) => fetchVaultPositions(vault, true),
+  });
+
+  await runInPromisePool({
+    items: Object.keys(vaultAssets),
+    concurrency: 5,
+    processor: (vault: string) => fetchVaultPositions(vault, false),
+  });
+
+  async function fetchVaultPositions(vault: string, isCurrent: boolean) {
```
g1nt0ki: should we have two flows, simpler one for vaults with less than 5M tvl or borrowed less than 80% and stricter check for the rest?

coins/src/adapters/moneyMarkets/morpho.ts
```diff
@@ -0,0 +1,277 @@
...
+
+async function morpho(
+  timestamp: number = 0,
+  vaultAssets: { [vault: string]: string } = {},
+  api: any,
+  target: string
+) {
+  const threeDaysAgo =
+    (timestamp == 0 ? getCurrentUnixTimestamp() : timestamp) - 3 * 24 * 60 * 60;
+  const threeDaysAgoBlock = await getBlock(api.chain, threeDaysAgo);
```
g1nt0ki: instead of getBlock, create another api, say `const apiThreeDaysAgo = new ChainApi(...)`

coins/src/adapters/oracles/pyth.ts
```diff
@@ -9,38 +9,62 @@ const now = getCurrentUnixTimestamp();
...
-    if (!id || !symbol || quote_currency !== 'USD') return;
-    feeds.push({ id: '0x' + id, symbol: symbol.replace(/\//g, ':') } as any);
-  })
+  feedsRes.forEach(
+    ({
+      id,
+      attributes: { symbol = undefined, quote_currency = undefined } = {},
+    }: any = {}) => {
+      if (!id || !symbol || quote_currency !== "USD") return;
+      feeds.push({ id: "0x" + id, symbol: symbol.replace(/\//g, ":") } as any);
```
0xpeluche: Replace it with this
```
const formattedSymbol = symbol.replace(/\//g, ":");
feeds.push({ id: "0x" + id, symbol: formattedSymbol, generic_symbol: formattedSymbol });
 ```
 because you're not assigning `generic_symbol` there, yet you're trying to access it later
 `const { symbol, generic_symbol } = feeds[idx];` → undefined issue

PR #11022, opened 2025-11-25
Fix: Update Titan Exchange website link to correct URL

0saurabh0:
updated the website and referral URL for Titan Exchange in the protocol data to use the correct link: https://titan.exchange. The previous value (titan.exchange/@defillama) was incorrect.


realdealshaman:
Hey @0saurabh0 fixed the url, thanks for the PR 🙏

PR #11024, opened 2025-11-26
euler bad debt

waynebruce0x:


coins/src/adapters/moneyMarkets/euler/eulerV2.ts
```diff
@@ -80,24 +89,56 @@ async function getEulerV2Tokens(
       chain,
       permitFailure: true,
     }),
-    sdk.api.abi.multiCall({
+    api.multiCall({
       calls: vaultAddresses.map((address: any) => ({
         target: address,
       })),
       abi: vaultAbi.find((m: any) => m.name === "totalAssets"),
       chain,
```
g1nt0ki: can skip the chain field for all of these

PR #11053, opened 2025-12-03
Add BMMF token mapping to tokenMapping.json

goktug-piku:
We want to add BMMF token to defilama price api


realdealshaman:
Hey @goktug-piku this token is already priced since it is listed on coingecko

PR #11129, opened 2025-12-18
Temp fix: reduce protocol route  response size

g1nt0ki:


noateden (dismissed):

defi/src/api2/utils/craftProtocolV2.ts
```diff
@@ -80,6 +94,11 @@ export async function craftProtocolV2({
     historicalTokenTvl = protocolCache[2] ?? []
   }
 
+  if (skipTokenBreakdownData) {
+    historicalUsdTokenTvl = []
+    historicalTokenTvl = []
```
0xpeluche: I’m wondering if we should secure the `skipTokenBreakdownData` path

Atm we set:
```
  historicalUsdTokenTvl = []
  historicalTokenTvl = []
```
later, in the `if (!useHourlyData)` append last record block, we can still `push(lastUsdTokenHourlyRecord)` / `push(lastTokenHourlyRecord)` because `[]` is truthy. This can re-inject 1 token breakdown datapoint, which feels inconsistent with the goal of not returning token breakdown ?

→
```
if (skipTokenBreakdownData) {
  historicalUsdTokenTvl = []
  historicalTokenTvl = []
  lastUsdTokenHourlyRecord = null
  lastTokenHourlyRecord = null
}
```

PR #11182, opened 2026-01-07
add har token metric for harmonix-finance adapter

vinhpad:


realdealshaman:
hey @vinhpad this PR isn't actually needed since we already pull the price of HAR from coingecko. I've also added it to the project listing to have the token details appear on the protocol's page

PR #11198, opened 2026-01-10
Fix tally proposals

FelixBruguera:
Continued from #11184


defi/src/governance/tally.ts
```diff
@@ -192,30 +207,30 @@ async function updateProposal(data: any, cache: any) {
...
-  const scores = [getVote('FOR'), getVote('AGAINST'), getVote('ABSTAIN'),]
+  const scores = [getVote('for'), getVote('against'), getVote('abstain'),]
   const scores_total = scores.reduce((acc, i) => acc + i, 0)
 
   let proposal = {
     ...data,
-    // start, end,
-    startBlock, endBlock, canceled, executed, eta,
+    start: start.number, 
+    end: end.number,
```
0xpeluche: `start/end` are now block numbers, but the rest of the pipeline treats them as unix timestamps → This will generate incorrect dates/stats

PR #11213, opened 2026-01-14
Rename 'StandX Perps' to 'StandX'

junlin-stx:
please also help update logo with
https://github.com/standx-labs/stdc_assets/blob/main/images/dusd.png

as it's missing right now.

image


realdealshaman:
@junlin-stx we have it set up that way for other perps chains as well, we need to keep it that way to be consistent

junlin-stx:
> What is your telegram handle, I can create the group

That would be great, thanks

My telegram handle:
@wjllance

PR #11396, opened 2026-02-14
Optimize coins chart route 

g1nt0ki:
fixes #11386

- if more than 500 data points are requested, throw an error
- if requested period is under 12h, use range query

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end tests for the coin price chart endpoint, including production-vs-local comparisons and a request-size limit test.

* **Performance**
  * Improved chart data retrieval with an adaptive hybrid query strategy to reduce latency and optimize efficiency across different time ranges.

* **Chores**
  * Disabled one batch historical span endpoint and added a test helper script to package tooling.

### DIFF
--- a/coins/serverless.yml
+++ b/coins/serverless.yml
@@ -181,13 +181,13 @@ functions:
       - httpApi:
           path: /batchHistorical
           method: get
-  batchHistoricalSpan:
-    handler: src/getBatchHistoricalCoinsSpan.default
-    timeout: 60
-    events:
-      - httpApi:
-          path: /batchHistoricalSpan
-          method: get
+  # batchHistoricalSpan: # not used by anyone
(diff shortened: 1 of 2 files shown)


0xngmi (commented):
have you done some test on whether using the limit:1 method could be always cheaper? maybe thats the case since ddb pricing is based on how much data you pull and this would lower the amount of data to its minimum

also it would be good to modify [getRecordClosestToTimestamp.ts](https://github.com/DefiLlama/defillama-server/blob/master/defi/src/utils/shared/getRecordClosestToTimestamp.ts) to always use the limit:1 method, just need to make sure it doesnt create issues anywhere else in the code

waynebruce0x (approved):

PR #11716, opened 2026-04-10
add lending & dex tags for omnipair

henrystats:
## Summary by CodeRabbit

* **Chores**
  * Updated protocol metadata: audit count set to 2 and added two audit links for Omnipair.

### DIFF
--- a/defi/src/protocols/data5.ts
+++ b/defi/src/protocols/data5.ts
@@ -14600,7 +14600,7 @@ const data5: Protocol[] = [
     audits: "0",
     gecko_id: "omnipair",
     cmcId: null,
-    tags: ["Dex", "Lending"],
+    tags: ["AMM", "Lending"],
     chains: ["Solana"],
     module: "omnipair/index.js",
     treasury: "omnipair.js",
--- a/defi/src/protocols/data5.ts
+++ b/defi/src/protocols/data5.ts
@@ -14600,7 +14600,7 @@ const data5: Protocol[] = [
     audits: "2",
     gecko_id: "omnipair",
     cmcId: null,
-    tags: ["AMM", "Lending"],
+    category: "Lending",
     chains: ["Solana"],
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey @henrystats, we can only have one category per listing. If the project has more than one product ie amm dex and lending, then we need to tracks the metrics for each separately. 

Please create an adapter that tracks info for the dex so that it is sepatate from lending, I'll then add a new listing and tie them together under a parent protocol listing like in the case of PancakeSwap: https://defillama.com/protocol/pancakeswap

image

elrakabawi:
Hey @realdealshaman, Omnipair is a new DeFi primitive that combines a DEX with lending, they're not separate products, it's not any sort of gimmick, it's literally a DEX that lends liquidity right from the DEX. 

I don't wanna waste your time, but please check this if you want more info: https://docs.omnipair.fi/technical-breakdown/generalized-automated-market-maker

Is it the case that the platform itself doesn't support two tags per project (programmatically?).

realdealshaman:
@elrakabawi I see, currently we can still only have one category, our backend is not built to handle two categories. Since the adapter also exports a borrowed amount, Lending is still the better of the two categories

elrakabawi:
@realdealshaman yeah it's a bit tough to choose due to the fact that it will be excluded from AMMs analysis/research, but I agree that DEX part is the tech (means), but lending is the goal (product). 

Can we have `category: Lending` but keep `tags: ["AMM", "Lending"]` as well for more exposure?

Thanks

realdealshaman:
@elrakabawi sorry we don't have our backend set up to handle this

elrakabawi:
@realdealshaman no worries, thanks for the explanation though

@henrystats then it make more sense to revert that change and get it back to `category: Lending` ser?

henrystats:
done ser @realdealshaman @elrakabawi!

PR #11750, opened 2026-04-15
Coins

waynebruce0x:
## Summary by CodeRabbit

* **New Features**
  * Added SAFO RWA adapter with Arbitrum oracle price feeds and DB write outputs
  * Extended Compound v3 market support to include an Ethereum comet
  * Added many new yield-farming tokens on Ethereum and Arbitrum
  * Added enzoBTC token mapping on Monad network


0xpeluche (approved):

g1nt0ki (changes requested):
two issues here:

 - not checking if oracle price is stale
 - should update price only on arbitrum, and map the same token on other chains to it, so we avoid redundant writes

PR #11848, opened 2026-04-28
Fix stale external response cache refresh

ametel01:
## Summary
- Fix stale cached external responses to invoke the refresh worker with the decoded url payload.
- Add unit coverage for cache miss, stale cache, and fresh cache paths.

## Why
Cache misses already invoked `cacheExternalResponse` with `{ url }`, but stale cache refreshes passed the full API Gateway event. The worker expects `event.url`, so stale entries logged `URL not defined` and skipped the background refresh.

## Test
- `cd defi && pnpm exec jest src/externalEndpointsCache/index.test.ts --runInBand`

## Release Notes

* **Tests**
  * Added comprehensive test coverage for external endpoints caching behavior, including scenarios for missing, stale, and fresh cached responses

* **Refactor**
  * Optimized the cache refresh mechanism for external endpoints


g1nt0ki:
@ametel01 unusure how it went unnoticed for so long, thanks for the bugfix

PR #11851, opened 2026-04-28
Complete per-chain TVL fallback (closes #11354)

### LINKED ISSUE #11354: add fallback feature - for tvl & dimensions data
add an option to use last working value max up to a week old if a smaller chain in a protocol breaks. For example, if a protocol with 100M in tvl breaks because of a chain but has only about 1k on that chain, doesnt make sense to fail the whole adapter, instead return cached response for the chain

track these failures and fix/remove the problematic chain from the adapter

DeFiDonkey:
Closes #11354.

## What this PR does

Completes the per-chain TVL fallback that's already 80% wired in `getAndStoreTvl.ts` but missing the **time bound** the issue specifically asks for ("max up to a week old"), and adds the **per-(protocol, chain) failure tracking** the issue's second sentence calls for ("track these failures and fix/remove the problematic chain").

## What's already in `master` (for context)

`defi/src/storeTvlInterval/getAndStoreTvl.ts` already:

- Captures per-chain failures via `tvlErrorsObject` (try/catch in `getTvl`).
- On cron-task success, writes a per-chain TVL snapshot to Redis (`storeTvlInterval-cache-${protocol.id}-${storedKey}`).
- On per-chain failure, calls `getCachedTvlData` to substitute the cached value if available.
- Has a 5%-of-total-TVL share threshold in `isCacheDataFresh`.

## What was missing

`isCacheDataFresh` (previously at lines 635–641) had an explicit comment: *"if the cached tvl is less than 5% of total tvl, consider it fresh **regardless of cache age**."* So a cached value from 6 months ago would still be substituted as long as it stayed below 5% — silently masking long-broken adapters. The issue text — *"max up to a week old"* — is exactly this missing constraint.

There was also no persistent record of which (protocol, chain) pairs were failing, no signal to the team that a chain has been broken for days, and the team's "fix/remove the problematic chain" workflow was unsupported.

## What this PR adds

1. **Time-bound check** in `isCacheDataFresh` (env-tunable, default 7 days via `TVL_FALLBACK_MAX_AGE_SEC`).
2. **Existing share threshold made env-tunable** (`TVL_FALLBACK_MAX_SHARE`, default `0.05`). Defaults are unchanged.
3. **`isCacheDataFresh` extracted to module-level + exported** so it's directly unit-testable. It also now returns a `reason` field (`'too-old' | 'over-threshold' | 'malformed-cache'`) that flows into the failure-tracking record.
4. **`throwWithReason` helper** so `getCachedTvlData` annotates re-thrown errors with the precise reason fallback was rejected.
5. **New module `defi/src/storeTvlInterval/chainTvlFailures.ts`** with three functions:
   - `recordChainFailure` — upserts into `chain_tvl_failures` (PRIMARY KEY `(protocol_id, chain)`, increments `consecutive_failures` on conflict). Wired into the failure block of `storeTvl`. Fail-soft: a Postgres outage never breaks the TVL pipeline.
   - `clearChainFailure` — deletes the row when the chain runs cleanly again. Wired into the success path of `storeTvl`.
   - `notifyChainTvlFailures` — daily summary to `STALE_COINS_ADAPTERS_WEBHOOK`; chains failing more than 72h are escalated to `TEAM_WEBHOOK`. Mirrors the structure of `notifyStaleCoins` so the operational pattern is familiar.
6. **Unit tests** (`getAndStoreTvl.test.ts`, 9 cases) covering: malformed cache, fresh cache, age-boundary, share-boundary, both env overrides (tighter + looser), and ordering when both checks fail.

## Behavior change

- Default behavior changes only at the limit: caches older than 7 days that *previously* would have been substituted (because they passed the 5% share check) will now be rejected with `reason: 'too-old'` and the adapter will fail loudly. This matches the issue's explicit ask.
- All other paths preserve current behavior.
- Both thresholds are env-tunable so the team can dial them per-environment without redeploying.

## New table required

```sql
CREATE TABLE IF NOT EXISTS chain_tvl_failures (
  protocol_id TEXT NOT NULL,
  chain TEXT NOT NULL,
  error_class TEXT,
  last_failure_at BIGINT NOT NULL,
  fallback_used BOOLEAN NOT NULL DEFAULT FALSE,
  fallback_reason TEXT,
  consecutive_failures INTEGER NOT NULL DEFAULT 1,
  cached_tvl NUMERIC NOT NULL DEFAULT 0,
  total_tvl NUMERIC NOT NULL DEFAULT 0,
  PRIMARY KEY (protocol_id, chain)
);
CREATE INDEX IF NOT EXISTS chain_tvl_failures_age ON chain_tvl_failures (last_failure_at);
```

I haven't included this as a migration file because the repo doesn't appear to have a schema-migration system for the existing `stalecoins` table — happy to follow whatever process you use for schema changes there.

## Files touched

- `defi/src/storeTvlInterval/getAndStoreTvl.ts` — extract + extend `isCacheDataFresh`, add `throwWithReason`, wire failure recording
- `defi/src/storeTvlInterval/chainTvlFailures.ts` — new module
- `defi/src/storeTvlInterval/getAndStoreTvl.test.ts` — new unit tests

## Verification

```bash
cd defi
npx jest src/storeTvlInterval/getAndStoreTvl.test.ts
# 9 passed
```

TypeScript: `npx tsc --noEmit` shows no new errors introduced by this PR (the pre-existing errors in `craftCsvDataset.ts` and elsewhere are unchanged).

## Notes / open questions

- I left the existing Elastic logging via `tempCacheInfo` (line 456-481) untouched — that's still useful for the "every-cache-use" stream. The new Postgres tracking complements it with a *persistent + queryable* record specifically for chains that need maintainer action.
- If you'd prefer the failure tracker behind a feature flag (e.g., `TVL_FAILURE_TRACKING_ENABLED`), happy to gate it. The current implementation is fail-soft so a missing table or DB outage can't break the pipeline, but I understand if you want to roll it out gradually.
- Defaults (7d age, 5% share) are conservative and align with the issue text. Easy to tune if production data suggests otherwise.

## Summary by CodeRabbit

* **New Features**
  * Per-chain TVL failure tracking with automated notifications and escalation for persistent failures (escalates after configured hours).
  * Cache-freshness checks for TVL fallbacks with configurable age and share thresholds; fallback usage and failure reasons are recorded and influence behavior.

* **Tests**
  * Added tests validating cache freshness logic, threshold behavior, env-configurable settings, and related edge cases.

### DIFF
--- a/defi/src/storeTvlInterval/chainTvlFailures.ts
+++ b/defi/src/storeTvlInterval/chainTvlFailures.ts
@@ -19,7 +19,8 @@ export type ChainTvlFailure = {
   protocol_id: string;
   chain: string;
   error_class: string;
-  last_failure_at: number;        // unix seconds
+  first_failure_at: number;       // unix seconds — preserved across upserts so escalation measures continuous failure duration
+  last_failure_at: number;        // unix seconds — bumped on every failure
   fallback_used: boolean;
   fallback_reason: string | null; // 'too-old' | 'over-threshold' | 'malformed-cache' | 'no-cache' | 'not-cron-task' | 'low-total-tvl' | null
   cached_tvl: number;
@@ -30,6 +31,7 @@ export const columns = [
   "protocol_id",
   "chain",
   "error_class",
+  "first_failure_at",
   "last_failure_at",
   "fallback_used",
   "fallback_reason",
@@ -48,6 +50,7 @@ export async function recordChainFailure(f: ChainTvlFailure): Promise<void> {
         on conflict (protocol_id, chain) do update set
           error_class = excluded.error_class,
           last_failure_at = excluded.last_failure_at,
+          -- first_failure_at deliberately NOT updated on conflict so escalation reflects continuous failure duration
           fallback_used = excluded.fallback_used,
           fallback_reason = excluded.fallback_reason,
           consecutive_failures = chain_tvl_failures.consecutive_failures + 1,
@@ -82,7 +85,7 @@ export async function notifyChainTvlFailures(): Promise<void> {
     const sql = await getPgConnection();
     const allColumns = [...columns, "consecutive_failures"];
     const stored: (ChainTvlFailure & { consecutive_failures: number })[] = await queryPostgresWithRetry(
-      sql`select ${(sql as any)(allColumns)} from chain_tvl_failures order by last_failure_at asc`,
+      sql`select ${(sql as any)(allColumns)} from chain_tvl_failures order by first_failure_at asc`,
       sql,
     );
     if (!stored.length) return;
@@ -91,7 +94,10 @@ export async function notifyChainTvlFailures(): Promise<void> {
     let summary = "";
     let escalation = "";
     for (const f of stored) {
-      const ageHours = (now - f.last_failure_at) / 3600;
+      // Age is measured from first_failure_at so escalation tracks continuous failure duration,
+      // not just time since the latest poke. A chain that fails every cron run for 5 days reports
+      // 5d here even though last_failure_at is "minutes ago".
+      const ageHours = (now - f.first_failure_at) / 3600;
       const ageDays = ageHours / 24;
       const fallbackInfo = f.fallback_used
         ? `fallback used (${f.fallback_reason ?? "ok"})`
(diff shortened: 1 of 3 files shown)


noateden:
hi @DeFiDonkey it's our internal task and I have already updated it

PR #11853, opened 2026-04-28
fix(security): harden internal debug cache routes

borngraced:
Currently, the internal debug cache routes only enforced authentication in the local-testing `API2_SKIP_SUBPATH === 'true'` path, which left the normal `API2_SUBPATH` deployment path without the intended secret check...the same handler also passed unsanitized cache keys into the file-backed PG cache helpers, which allowed traversal-style keys to escape the `pg-cache` directory. 
Now, dedicated internal secret for all debug cache operations is required, rejects invalid cache keys before any cache read/write/delete, and adds regression tests covering unauthenticated access, authenticated access, and traversal attempts.

## Release Notes

* **Bug Fixes**
  * Strengthened cache key validation with normalization to prevent unauthorized directory traversal attacks.

* **Tests**
  * Added comprehensive test coverage for internal debug route authentication and cache key validation security.

### DIFF
--- a/defi/src/api2/cache/file-cache.ts
+++ b/defi/src/api2/cache/file-cache.ts
@@ -126,7 +126,31 @@ function getCacheFile(key: string) {
   return `pg-cache/${key}`
 }
 
+const INVALID_CACHE_KEY_ERROR = 'Invalid cache key'
+
+function normalizePGCacheKey(key: string) {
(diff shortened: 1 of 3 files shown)


g1nt0ki (commented):
@borngraced thanks for the PR, I fixed it, we ensure that we always have the api key, and I have nuked the read pg route and other code that we no longer use

PR #11859, opened 2026-04-28
fix(getProtocolTvl): correct allTvlsAreAddl aggregation; extract testable helper

AimenHallou:
## Summary

`getProtocolTvl.ts` has a small but real logic bug at lines 184-191 of `defi/src/utils/getProtocolTvl.ts`:

```ts
let allTvlsAreAddl = false;

Object.keys(chainTvls).forEach((type) => {
  allTvlsAreAddl =
    type === "doublecounted" ||
    type === "liquidstaking" ||
    type === "dcAndLsOverlap";
});
```

The `forEach` callback assigns to `allTvlsAreAddl` on every iteration instead of aggregating, so after the loop the value reflects only the LAST key in `chainTvls`, not whether ALL keys are additional aggregates. When `chainTvls` contains a real chain key plus an additional-aggregate key (e.g. `{ Ethereum, doublecounted }`) and the additional one happens to be last in iteration order, the flag flips to `true` and triggers the `chainsLength <= 3 && allTvlsAreAddl` fallback that backfills the protocol's default chain TVL — double-counting in that fallback path.

## Fix

Extract a top-level `allKeysAreAdditionalTvl(chainTvls)` helper that does what the variable name claims (every key must be one of the three additional-aggregate keys), and replace the inline loop with a single call.

## Test

New `defi/src/utils/getProtocolTvl.test.ts` covers the four regression cases:

- Empty record -> `false`
- All additional keys -> `true`
- Mixed (real chain + additional) -> `false` (the case the previous bug got wrong)
- Single real chain key -> `false`

Runs with `npx jest src/utils/getProtocolTvl.test.ts --no-coverage` from `defi/`.

## Risk

Behavior moves toward the variable's stated intent. No public API change. Existing callers continue to work because the only call site is the inline loop being replaced.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for protocol total value locked calculations with extensive coverage of edge cases and default chain fallback behaviors.

* **Refactor**
  * Improved code organization and maintainability through extraction of a dedicated helper function for TVL category validation and detection logic.

### DIFF
--- a/defi/src/utils/getProtocolTvl.test.ts
+++ b/defi/src/utils/getProtocolTvl.test.ts
@@ -1,0 +1,24 @@
+import { allKeysAreAdditionalTvl } from "./getProtocolTvl";
+
+describe("allKeysAreAdditionalTvl", () => {
+  it("returns false for an empty record", () => {
+    expect(allKeysAreAdditionalTvl({})).toBe(false);
+  });
+
+  it("returns true when every key is an additional aggregate", () => {
+    expect(allKeysAreAdditionalTvl({ doublecounted: {}, liquidstaking: {} })).toBe(true);
+    expect(allKeysAreAdditionalTvl({ dcAndLsOverlap: {} })).toBe(true);
+    expect(
+      allKeysAreAdditionalTvl({ doublecounted: {}, liquidstaking: {}, dcAndLsOverlap: {} })
+    ).toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


AimenHallou:
@coderabbitai review

AimenHallou:
@coderabbitai review

AimenHallou:
@coderabbitai full review

PR #11867, opened 2026-04-28
add usdnr and M prices on fluent

RohanNero:
## Summary by CodeRabbit

* **New Features**
  * Added USDnr token support on the Fluent chain with per-token price calculations.
  * Updated token mappings for the citrea network and introduced new Fluent network mapping with M token configuration.


RohanNero:
#11881

PR #11868, opened 2026-04-28
 Move misrepresented token lookup down in token search as importAdapter is sync

ratik:
### Summary
* change _protocolHasMisrepresentedTokens to sync as importAdapter is sync as well
*  Move misrepresented token lookup down in token search as importAdapter is sync and there is possible early exit

## Summary by CodeRabbit

* **Refactor**
  * Protocol validation now checks TVL shape and match count first; token misrepresentation checks run only for protocols that pass these validations. Misrepresentation checks were changed from asynchronous/concurrent evaluation to synchronous/sequential evaluation and now return an explicit boolean result, yielding more predictable and consistent validation behavior.


ratik:
@coderabbitai review

PR #11883, opened 2026-04-29
fix: stop api test fallthrough

NFToby:
add missing break after 'api-test-stop' to prevent fallthrough, which could incorrectly trigger spike command when stopping API tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where stopping API tests could inadvertently trigger unintended operations.


g1nt0ki (approved):
nice catch

PR #11886, opened 2026-04-30
fix(coins/coingeckoPlatforms): preserve leading character in padAddress when input has no 0x prefix

AimenHallou:
## Summary

`padAddress` in `coins/src/utils/coingeckoPlatforms.ts` (lines 33-39) has a small but real bug when the input lacks a `0x` prefix:

```ts
const data = address.substring(address.indexOf(prefix) + prefix.length);
```

`indexOf("0x")` returns `-1` when the prefix is absent, so this becomes `substring(-1 + 2)` = `substring(1)`, silently dropping the first character. `padAddress("1234", 8)` returns `"0x000234"` (missing the leading `1`) instead of `"0x001234"`.

## Fix

```ts
const data = address.startsWith(prefix) ? address.slice(prefix.length) : address;
```

This treats a missing prefix as "the whole input is the data part" rather than chopping a character off.

## Tests

Adds a `describe('padAddress', ...)` block to the existing `coingeckoPlatforms.test.ts` covering five cases: with-prefix, without-prefix (the regression case for this bug), empty input, input that is just the prefix, and the default-length path. Run with `npx jest src/utils/coingeckoPlatforms.test.ts --no-coverage` from `coins/`.

## Honest framing

This is a latent bug. Every current caller (`lowercase()` on the starknet path, the `processCoin` and `getWrites` flows that feed it) passes addresses that already start with `0x`, so the bug is invisible in production today. This PR is defensive hardening: fix the helper so it does what its name says regardless of input shape, and lock the corrected behavior with tests so future callers can rely on it without rediscovering the edge case. No public API change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed address processing logic to correctly handle addresses with different formatting variations, preventing errors when addresses lack expected prefix patterns.

* **Tests**
  * Added comprehensive test coverage for address validation, including edge cases for multiple input formats and prefix scenarios.

### DIFF
--- a/coins/src/utils/coingeckoPlatforms.ts
+++ b/coins/src/utils/coingeckoPlatforms.ts
@@ -32,7 +32,7 @@ export interface CoinMetadata {
 
 export function padAddress(address: string, length: number = 66): string {
   let prefix = "0x";
-  const data = address.substring(address.indexOf(prefix) + prefix.length);
+  const data = address.startsWith(prefix) ? address.slice(prefix.length) : address;
   const zeros = length - prefix.length - data.length;
   for (let i = 0; i < zeros; i++) prefix += "0";
   return prefix + data;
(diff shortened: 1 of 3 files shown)


AimenHallou:
@coderabbitai review

PR #11954, opened 2026-05-08
Secure API2 internal debug-pg cache routes

### LINKED ISSUE #11953: Secure API2 internal debug-pg cache routes
## Summary

The API2 internal `debug-pg` routes are mounted under `/<API2_SUBPATH>_internal/debug-pg/*`, but the internal secret check currently only runs when `API2_SKIP_SUBPATH === 'true'`. That appears to be the local/test path, so production internal routes can read from and delete the PG/file cache without requiring `x-internal-secret`.

## Impact

Unauthenticated callers who know or discover the API2 internal subpath can:

- Read cached API2 data through `GET /<API2_SUBPATH>_internal/debug-pg/*`.
- Delete cached API2 data through `DELETE /<API2_SUBPATH>_internal/debug-pg/*`.
- Trigger `clearDimensionsCacheV2()` through the special `DELETE /clear-dimensions-cache` path.

The same path is forwarded into cache helpers, so the fix should also reject traversal-like or absolute cache keys before any file-cache operation.

## Affected code

- `defi/src/api2/index.ts` mounts the internal router when `API2_SUBPATH` is set.
- `defi/src/api2/routes/internalRoutes.ts` registers `GET` and `DELETE` for `/debug-pg/*`.
- `defi/src/api2/routes/internalRoutes.ts` only checks `x-internal-secret` when `API2_SKIP_SUBPATH === 'true'`.
- `defi/src/api2/cache/file-cache.ts` maps cache keys onto local files.

## Proposed fix

- Require `x-internal-secret` for `debug-pg` routes in every environment.
- Use `LLAMA_INTERNAL_ROUTE_KEY` as the required secret and fail closed if it is unset.
- Do not fall back to public/pro route values such as `API2_SUBPATH`.
- Validate debug cache keys before passing them to `readFromPGCache` or `deleteFromPGCache`:
  - reject empty keys
  - reject absolute paths
  - reject `..` traversal segments
  - normalize to a slash-prefixed cache key for existing cache helper compatibility
- Return `401`/`403` for auth failures and `400` for unsafe cache paths, instead of reporting everything as a 500.
- Add focused tests for auth and path validation.

ametel01:
Fixes #11953

## Summary
- require `LLAMA_INTERNAL_ROUTE_KEY` for API2 internal `debug-pg` cache deletion endpoints
- remove the stale `GET /debug-pg/*` registration after upstream removed the cache read handler
- reject missing/misconfigured internal route secrets with explicit error responses
- normalize and validate debug cache keys before deleting cache entries
- send the internal secret from protocol cache clearing utilities, including dimensions cache clearing
- add regression coverage for auth outcomes and cache key traversal rejection

## Security notes
The affected internal route allowed cache mutation under `/<API2_SUBPATH>_internal/debug-pg/*` when reachable. The fix makes these operations require `x-internal-secret: $LLAMA_INTERNAL_ROUTE_KEY` and rejects malformed, empty, root, and traversal-style cache paths. It also removes fallback secrets from `LLAMA_PRO_API2_SECRET_KEY` and `API2_SUBPATH`.

## Conflict resolution
Merged current `DefiLlama/master` into the branch and resolved the conflict in `defi/src/api2/routes/internalRoutes.ts` by keeping upstream's deprecated-code removal while preserving this PR's auth and path-validation checks.

## Verification
- `pnpm exec jest src/api2/routes/internalRouteSecurity.test.ts --runInBand`
- `pnpm exec tsc --noEmit --pretty false 2>&1 | rg 'src/(api2/routes/internalRoute|cli/utils/clearProtocolCache)|internalRoutes\\.ts' || true`

Full `pnpm exec tsc --noEmit --pretty false` previously reported existing repo-wide TypeScript errors outside this patch, including generated Smithy duplicate exports, missing dimension adapter modules, and pre-existing unused variables in API tests.


g1nt0ki (changes requested):
sorry, rejecting the attack vector is person knowing api secret - if someone finds this, we have bigger problems compared to them deleting cache files

PR #11978, opened 2026-05-11
Add zero-value dimensions refill option

Genmin:
## Summary

- adds a `refill_zero_value_data=true` mode to the dimensions refill script for records that exist in DB but only contain zero values
- supports optional `zero_record_keys=df,dr` filtering so operators can target the affected aggregate keys for fees/revenue backfills
- keeps `dry_run=true` as the default and prints the selected zero-value dates before running refill work

## Context

This is meant for cases like Renzo fees after 2026-03-06 where rows already exist, but they contain zero data. `refill_only_missing_data=true` skips those days because the records are present, so operators need a way to refill present-but-zero records without hard-coding a protocol-specific script.

Example dry run for the Renzo fees issue:

```sh
type=fees protocol=renzo from=2026-03-07 to=2026-05-10 refill_zero_value_data=true zero_record_keys=df,dr dry_run=true npm run fillOld-dimensions
```

## Validation

- `npm install --legacy-peer-deps --no-package-lock --no-audit --no-fund`
- `npm run prebuild-gen-imports`
- `npm run load-dimensions-repo --silent`
- `npx tsc --noEmit --pretty false --skipLibCheck --noUnusedLocals false --noUnusedParameters false --target ES2020 --module commonjs --esModuleInterop --resolveJsonModule src/adaptors/handlers/storeAdaptorData/refillScript.ts`
- `git diff --check`

Closes #11969.

## New Features

* Added support for refilling zero-value dimension records with a new configurable mode.
* Runtime configuration now reports zero-value mode status and record key counts for improved visibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/defillama-server/pull/11978)


bheluga:
@Genmin thanks for the PR
We already refill zero vaules, the problem here was due to a bug in dimension adapter

